PR #4, opened 2017-03-16
Allow sssd responders to run as socket activated services

lslebodn:
https://bugzilla.redhat.com/show_bug.cgi?id=1416318#c1

Please provide feedback to this trivial solution of RHBZ1416318. All responder should be running with `sssd_t` domain but they do not require all privileges. I have a plan to run responders with different domain  will less privileges.


lslebodn:
Bump

PR #7, opened 2017-04-04
Remove ejabberd from the RabbitMQ policy and add an ejabberd policy

bowlofeggs:
This PR has two commits. The first one stops the rabbitmq policy from matching ejabberd, and the second adds an ejabberd policy.

https://bugzilla.redhat.com/show_bug.cgi?id=1424823

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


lemenkov:
Step in the right direction!

wrabcak:
ejabberd policy was backported from rawhide branch.

PR #10, opened 2017-04-22
 Add a boolean to enable the use of dac_override

mscherer:
This is needed for tor onion services, as seen on
https://bugzilla.redhat.com/show_bug.cgi?id=1392187


mscherer:
That is a copy of a old PR that was closed without message

PR #12, opened 2017-05-17
Use stricter fc rules for sssd sockets in /var/run

lslebodn:
* mark them as sockets
* escape dots

sh$ semanage fcontext -l | grep heim_org
/var/run/\.heim_org\.h5l\.kcm-socket socket system_u:object_r:sssd_var_run_t:s0

### DIFF
--- a/sssd.fc
+++ b/sssd.fc
@@ -26,5 +26,5 @@
 /var/log/sssd(/.*)?		gen_context(system_u:object_r:sssd_var_log_t,s0)
 
 /var/run/sssd.pid	--	gen_context(system_u:object_r:sssd_var_run_t,s0)
-/var/run/secrets.socket		gen_context(system_u:object_r:sssd_var_run_t,s0)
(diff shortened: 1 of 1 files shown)


sssd.fc
```diff
@@ -26,5 +26,5 @@
 /var/log/sssd(/.*)?		gen_context(system_u:object_r:sssd_var_log_t,s0)
 
 /var/run/sssd.pid	--	gen_context(system_u:object_r:sssd_var_run_t,s0)
-/var/run/secrets.socket		gen_context(system_u:object_r:sssd_var_run_t,s0)
-/var/run/.heim_org.h5l.kcm-socket	--  	gen_context(system_u:object_r:sssd_var_run_t,s0)
+/var/run/secrets\.socket	-s	gen_context(system_u:object_r:sssd_var_run_t,s0)
```
wrabcak, line 29: LGTM. 
These files are sockets.

PR #15, opened 2017-05-26
[RFE] Rabbitmq logrotate

lemenkov:
Enhancements, which allows logrotate to rotate RabbitMQ logs properly.

!!! PLEASE DO NOT MERGE - this is still a w.i.p.

PR #16, opened 2017-05-29
Split-off EPMD-related rules

lemenkov:
Signed-off-by: Peter Lemenkov <lemenkov@redhat.com>


epmd.if
```diff
@@ -0,0 +1,12 @@
+interface(`epmd_query',`
```
wrabcak, line 1: I don't understand why this change is needed.

ejabberd.te
```diff
@@ -31,12 +31,12 @@ auth_use_nsswitch(ejabberd_t)
 corecmd_exec_bin(ejabberd_t)
 corecmd_exec_shell(ejabberd_t)
 
-corenet_tcp_bind_epmd_port(ejabberd_t)
+epmd_query(rabbitmq_t)
```
wrabcak: You have here rabbitmq_t domain, which mean that you allowing rules in this macro for rabbitmq_t instead of ejabberd_t, but I still don't understand why you create epmd_query() interface. Do you have some BZ for this?

wrabcak (changes requested):

lemenkov:
@wrabcak fixed ejabberd_t / rabbitmq_t typo. Thanks for pointing me on this.

Regarding your question about EPMD rules splitting off. It's a very generic application and the corresponding workflow is very similar for many other Erlang applications which require clustering. So better to split it off and reuse in dedicated *.te files, as I did for Ejabberd and RabbitMQ.

Also any *ctl scripts for these native Erlang clustering application will likely require interaction with EPMD as well (think of monitoring, logs rotation, and any other similar operations), so we'll need to add necessary rules as well. To avoid duplication we'd better to split off EPMD operations into its own  generic interface (epmd_query).

Some Erlang applications won't require it. For example, CouchDB doesn't need it because it doesn't use native Erlang clustering.

wrabcak:
Any update here?

fpoirotte:
Hi,

Any update on this PR & #15 ?

To give some context, these PR are meant to fix https://bugzilla.redhat.com/show_bug.cgi?id=1413775, which I'm currently facing as well. I don't know much about SELinux, but what would be needed to get these approved?

Regards

PR #24, opened 2017-08-17
dbus: add policy for dbus-broker

teg:
dbus-broker is a drop in replacement for dbus-daemon. It can therefore
mostly simply rely on the existing dbus policy module. However, it also
needs to have its binaries labeled correctly, and it needs permission to
perform the D-Bus method call StartTransientUnit on PID1, which
dbus-daemon did not.

For details see <https://github.com/bus1/dbus-broker/wiki>.

Package submission to rawhide: <https://bugzilla.redhat.com/show_bug.cgi?id=1482202>.
Merge into refpolicy: <https://github.com/TresysTechnology/refpolicy-contrib/pull/64>.

### DIFF
--- a/dbus.fc
+++ b/dbus.fc
@@ -9,6 +9,8 @@ ifdef(`distro_redhat',`
 ')
 
 /usr/bin/dbus-daemon(-1)? --	gen_context(system_u:object_r:dbusd_exec_t,s0)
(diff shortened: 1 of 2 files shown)


wrabcak:
Thanks.

PR #26, opened 2017-08-31
Add initial usbguard policy

vinzent:
The USBGuard software framework helps to protect your computer
against rogue USB devices.


vinzent:
problems:

* usbguard opens usbguard-daemon.conf  in read/write mode or it fails - for servers i'd expect this to be undesirable as I would manage the config file by some config management tool and not let usbguard to manage the config itself

dkopecek:
@vinzent FYI the next release will introduce /var/run/usbguard.pid and a double fork daemonization procedure (systemd unit will change from Type=Simple to Type=Forking). See https://github.com/dkopecek/usbguard/commit/e4f8b6ae0b51a181cd27c59121c2f52e0305e10a#diff-d64ca99ed9d6a2359094e311a9bb375bR8

vinzent:
@dkopecek is there any RPM available which is built from master?

dkopecek:
@vinzent try this: https://copr.fedorainfracloud.org/coprs/mildew/usbguard-testing/

usbguard.fc
```diff
@@ -0,0 +1,7 @@
+/etc/usbguard/rules.conf                  -- gen_context(system_u:object_r:usbguard_rules_t,s0)
+/etc/usbguard(/.*)?                          gen_context(system_u:object_r:usbguard_conf_t,s0)
+/dev/shm/qb-usbguard-.*                   -- gen_context(system_u:object_r:usbguard_tmpfs_t,s0)
+/usr/bin/usbguard                         -- gen_context(system_u:object_r:usbguard_exec_t,s0)
+/usr/lib/systemd/system/usbguard.service  -- gen_context(system_u:object_r:usbguard_daemon_unit_file_t,s0)
```
wrabcak: I prefer /usr/lib/systemd/system/usbguard.* here because of future systemd unit files.

usbguard.te
```diff
@@ -0,0 +1,79 @@
...
+## <p>
+## Allow usbguard-daemon to write rules (e.g. /etc/usbguard/rules.conf)
+## </p>
+## </desc>
+gen_tunable(usbguard_daemon_write_rules, false)
+
+type usbguard_daemon_t;
+type usbguard_daemon_exec_t;
+init_daemon_domain(usbguard_daemon_t, usbguard_daemon_exec_t)
+permissive usbguard_daemon_t;
```
wrabcak: This should be removed before we add this to distribution policy.

usbguard.te
```diff
@@ -0,0 +1,79 @@
...
+
+type usbguard_daemon_initrc_exec_t;
```
wrabcak: We don't need this label for usbguard. This can be removed.

usbguard.te
```diff
@@ -0,0 +1,79 @@
...
+init_script_file(usbguard_daemon_initrc_exec_t)
+
+type usbguard_daemon_unit_file_t;
+systemd_unit_file(usbguard_daemon_unit_file_t)
+
+type usbguard_t;
```
wrabcak: @dkopecek ,
How is /usr/bin/usbguard  executed? I expect by user on command line or some gui tool?  If yes, in most cases this will run as unconfined_t domain.

usbguard.te
```diff
@@ -0,0 +1,79 @@
...
+dev_list_sysfs(usbguard_daemon_t)
+dev_rw_sysfs(usbguard_daemon_t)
+
+read_files_pattern(usbguard_daemon_t,usbguard_conf_t,usbguard_conf_t)
+read_files_pattern(usbguard_daemon_t,usbguard_conf_t,usbguard_rules_t)
+
+manage_files_pattern(usbguard_daemon_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+fs_tmpfs_filetrans(usbguard_daemon_t, usbguard_tmpfs_t, file)
+
+manage_files_pattern(usbguard_daemon_t, usbguard_log_t, usbguard_log_t)
```
wrabcak: Also, managing for dirs could be added:
manage_dirs_pattern(usbguard_daemon_t, usbguard_log_t, usbguard_log_t)

usbguard.te
dkopecek: Yes, it is a CLI tool used by users.

usbguard.fc
vinzent: i'll change this.

usbguard.te
vinzent: indeed. :-)

usbguard.te
vinzent: ok. I'll remove it.

usbguard.te
vinzent: @wrabcak you commented on an old version of the PR (don't know but every now and then GH requires a page refresh or it shows stale content). i've already removed the /usr/bin/usbguard and renamed usbguard_daemon_t to usbguard_t.

@dkopecek  @wrabcak  the usbguard cli tool IMHO connects to the usbguard-daemon by sockets in shared memory ? limiting the access should be done on these sockets. This is not part of this policy upto now.  I've only run the usbguard as unconfined_t.  I think a interface and role would be required to support confined access.

usbguard.te
vinzent: @wrabcak the file is included in the RPM (@dkopecek but no logrotate config, i think this should be added or the file will grow forever ?) - this is why i've not added the manage dirs. IMHO the "safest" way would to have append only, no? so the daemon can't tamper the logs - but this would require support of the daemon.

vinzent:
@dkopecek there us a usbguard-dbus package. I think this needs some selinux policy too as dbus things are confined. 

how can this dbus thing be used?

usbguard.te
dkopecek: > @dkopecek @wrabcak the usbguard cli tool IMHO connects to the usbguard-daemon by sockets in > shared memory ? > limiting the access should be done on these sockets.

That is correct. There is some UID/GID checking done by usbguard-daemon on the application level. If we can limit access to the socket to the smallest possible set of users / processes that need it, then I say go for it :-)

dkopecek:
@vinzent usbguard-dbus exposes the usbguard IPC bus via D-Bus. It connects as an IPC client to usbguard-daemon and implements/exposes a D-Bus interfaces for D-Bus clients.

vinzent:
@wrabcak @dkopecek 

* removed permissive attribute
* changed systemd unit file context 
* removed unconfined_initrc_exec_t
* allowed  netlink_kobject_uevent_socket read

vinzent:
@wrabcak @dkopecek 

* added usbguard-dbus types
* added usbguard_ipc_access interface

question: is /var/run/usbguard.pid the correct path? /var/run is a symlink to /run

usbguard.fc
```diff
@@ -0,0 +1,8 @@
+/etc/usbguard/rules.conf                  -- gen_context(system_u:object_r:usbguard_rules_t,s0)
+/etc/usbguard(/.*)?                          gen_context(system_u:object_r:usbguard_conf_t,s0)
+/dev/shm/qb-usbguard-.*                   -- gen_context(system_u:object_r:usbguard_tmpfs_t,s0)
+/usr/lib/systemd/system/usbguard.*        -- gen_context(system_u:object_r:usbguard_unit_file_t,s0)
+/usr/sbin/usbguard-daemon                 -- gen_context(system_u:object_r:usbguard_exec_t,s0)
+/usr/sbin/usbguard-dbus                   -- gen_context(system_u:object_r:usbguard_dbus_exec_t,s0)
+/var/log/usbguard(/.*)?                      gen_context(system_u:object_r:usbguard_log_t,s0)
+/var/run/usbguard.pid                     -- gen_context(system_u:object_r:usbguard_var_run_t,s0)
```
wrabcak: Could be replaced by /var/run/usbguard.*

usbguard.te
```diff
@@ -0,0 +1,91 @@
...
+type usbguard_rules_t;
+files_config_file(usbguard_rules_t)
+
+type usbguard_tmpfs_t;
+files_tmpfs_file(usbguard_tmpfs_t)
+
+type usbguard_var_run_t;
+files_pid_file(usbguard_var_run_t)
+
+type usbguard_dbus_t;
```
wrabcak: This is not needed, we can use only usbguard_t domain.

vinzent: @wrabcak if I remove usbguard_dbus_t (and the `dbus_system_domain(usbguard_dbus_t, usbguard_dbus_exec_t)` 

needed to re-add `dbus_system_domain(usbguard_t, usbguard_exec_t)` 

why not have a dedicated type for the dedicated usbguard-dbus daemon process?

bachradsusi:
Tried on Fedora 27 without unconfined module and there are few rules missing:

    allow staff_t usbguard_t:unix_stream_socket connectto;
    allow staff_t usbguard_tmpfs_t:file { map open read write };
    allow usbguard_t usbguard_exec_t:file map;
    allow usbguard_t usbguard_tmpfs_t:file map;

vinzent:
> Tried on Fedora 27 without unconfined module and there
>   are few rules missing:

Trying to download F27 install media... seems like it takes some time to
load.


> |allow staff_t usbguard_t:unix_stream_socket connectto; allow staff_t
> usbguard_tmpfs_t:file { map open read write };|
whats this staff_t domain - what runs as staff_t type?

usbguard_ipc_access(staff_t)  should do it.

> |allow usbguard_t usbguard_exec_t:file map; allow usbguard_t
> usbguard_tmpfs_t:file map;|
I need to add the map perm in the usbguard_ipc_access() interface.

> |allow usbguard_t var_log_t:file open; |

var_log_t is suspicious. Log file is
/var/log/usbguard/usbguard-daemon.log - both should have the
usbguard_log_t types. What files were accessed ?

bachradsusi:
I edited my comment and dropped the log part as it was a labelling problem on my system.

> whats this staff_t domain - what runs as staff_t type?

staff_t type is assigned to staff_u which is one of so called confined SELinux users. I don't use unconfined_u for the user I use. More about my configuration can be found at https://plautrba.fedorapeople.org/selinux-confined-system-with-fedora-27.html

bachradsusi:
and https://docs-old.fedoraproject.org/en-US/Fedora/25/html/SELinux_Users_and_Administrators_Guide/chap-Security-Enhanced_Linux-Confining_Users.html

vinzent:
> I edited my comment and dropped the log part as it was a labelling
> problem on my system.
>
>     whats this staff_t domain - what runs as staff_t type?
>
> staff_t type is assigned to staff_u which is one of so called confined
> SELinux users. I don't use unconfined_u for the user I use. More about
> my configuration can be found at
> https://plautrba.fedorapeople.org/selinux-confined-system-with-fedora-27.html
>
>
i suspect this usbguard_ipc_access() should be put optionally in
selinux-policy.git/policy/modules/roles/sysadm.te and
selinux-policy.git/policy/modules/roles/staff.te  (sysadm should IMHO
also be able to set rules). At least I can't find staff_t references in
selinux-policy-contrib.git

dkopecek:
@vinzent I would like to ship the policy as part of the package. Would it be ok for you if I moved this contribution to https://github.com/USBGuard/usbguard-selinux? In the future I'd like to hook CI to this repository so that the policy will be tested on changes or new distribution releases.

wrabcak:
@dkopecek  +1

dkopecek:
@vinzent ping

wrabcak:
Any update here?

wrabcak:
Hi All, 
I'm closing this PR because of https://github.com/USBGuard/usbguard/issues/26 and https://github.com/USBGuard/usbguard-selinux/pull/1

@vinzent,
Thanks for your contribution!

PR #27, opened 2017-09-11
Allow pki_tomcat_t use nsswitch

lslebodn:
Partially related to ade89ff23d541a761fce0c4a3e6ba14d00ed9be7

I cannot see AVC on f25 only on f26 (and it's already fixed in f27)


wrabcak:
Thanks for PR

PR #28, opened 2017-09-12
Revert "Add rules fixing installing ipa-server-install with SELinux i…

lslebodn:
…n Enforcing. BZ(1488404)"

This reverts commit 1e84030553d5c8345344ad2f4e3e04062239c165
without reverting changes in rpm.te

It was but in ipa-server-install script
https://bugzilla.redhat.com/show_bug.cgi?id=1490762


wrabcak:
Thanks Lukas.

PR #29, opened 2017-09-18
Allow virtual machines to access TPM devices. BZ(1492635)

martinezjavier:
QEMU needs read and write access to TPM devices to expose them to guest machines using the pass-through backend.

https://bugzilla.redhat.com/show_bug.cgi?id=1492635


wrabcak:
Thanks.

PR #30, opened 2017-09-28
Ganesha is an nfs server, It should be running as nfsd_exec_t

rhatdan:
This policy moves the definition of ganesha out of glusterd into rpc.fc
Ganesha should follow the nfs booleans and be labeled correctly.


wrabcak:
Dan, 
I created scratch build for Fedora Rawhide where ganesha has nfsd_exec_t label, you should have it in mailbox. If we change the label we don't need this patch. Closing.

PR #33, opened 2017-10-11
Fix OpenVSwitch 2.8

lkundrak:
Tested with Fedora 27.

### DIFF
--- a/openvswitch.te
+++ b/openvswitch.te
@@ -32,7 +32,7 @@ systemd_unit_file(openvswitch_unit_file_t)
 # openvswitch local policy
 #
 
-allow openvswitch_t self:capability { net_admin ipc_lock sys_module sys_nice sys_rawio sys_resource chown setgid setpcap setuid };
+allow openvswitch_t self:capability { dac_override net_admin ipc_lock sys_module sys_nice sys_rawio sys_resource chown setgid setpcap setuid };
 allow openvswitch_t self:capability2 block_suspend;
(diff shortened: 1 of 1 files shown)


apevec:
audit/runuser is OVS 2.8.1 diff to OVS 2.7.3 in ovsdb-server.service:
```
+ExecStartPre=/usr/bin/chown ${OVS_USER_ID} /var/run/openvswitch
 ExecStart=/usr/share/openvswitch/scripts/ovs-ctl \
           --no-ovs-vswitchd --no-monitor --system-id=random \
+          --ovs-user=${OVS_USER_ID} \
           start $OPTIONS
 ExecStop=/usr/share/openvswitch/scripts/ovs-ctl --no-ovs-vswitchd stop
 ExecReload=/usr/share/openvswitch/scripts/ovs-ctl --no-ovs-vswitchd \
+           --ovs-user=${OVS_USER_ID} \
            --no-monitor restart $OPTIONS

```
introduced in https://github.com/openvswitch/ovs/commit/92d53574d5039e1173347754090cf64ccf2af57c

I'd ask OVS team why runuser could not be replaced with systemd User= option, I don't think running user needs to be configurable in the distro packaging.

wrabcak:
LGTM. 
@lkundrak , maybe we can avoid writing to audit log using idea from @apevec . If OVS team will change it we should remove the rule from SELinux policy.

apconole:
@apvec - there are a number of capabilities that OvS configures for itself, and needs to start as root then downgrade.

PR #34, opened 2017-10-11
Fix OpenVSwitch 2.8

lkundrak:
Opening yet another pull request, because no. 33 seems to have slipped through the cracks in GitHub and I can't access it anymore.

Tested on Fedora 27 (but the last commit).


lkundrak:
Closing it. It exist redundantly only due to GitHub hiccup.

PR #35, opened 2017-10-14
Allow postfix_master_t to bind to unreserved TCP ports. BZ(1373939)

aleksandrs-ledovskis:
This reanimates and fixes _WONTFIX_'ed BZ [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1373939). 
I have verified the fix by requesting password recovery for bugzilla.redhat.com - previously it failed due to `redhat.com` TXT record being larger than 512 bytes/requiring TCP re-query.


wrabcak:
LGTM.

PR #38, opened 2017-11-08
Remove the ejabberd policies.

bowlofeggs:
The ejabberd policies are now packaged by the ejabberd spec file,
as of the following commit:

https://src.fedoraproject.org/rpms/ejabberd/c/e9cdddd835473981e90316f35ac24247215d6313?branch=master

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #39, opened 2017-11-08
Remove the ejabberd policies.

bowlofeggs:
The ejabberd policies are now packaged by the ejabberd spec file,
as of the following commit:

https://src.fedoraproject.org/rpms/ejabberd/c/8216c0b32cf311e4ccf5e05cbe49c1f99bfacdf3?branch=f27

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>


wrabcak:
We should remove it in Rawhide first. Will close this task and continue in #38

wrabcak:
Hi @bowlofeggs , 
Sorry for really late reply. I tried package in fedora "ejabberd-selinux". It working well, however you could use our selinux-policy rpm macros to have spec file more clear. More info here[1]

There is also just one interface inside the ejabberd.if file, so we could remove it from Distribution policy. But we need to agreed that I'll remove it from selinux-policy and you'll add dependency in ejabberd-selinux package to require version of selinux-policy without the ejabberd policy. 

[1] https://fedoraproject.org/wiki/SELinux/IndependentPolicy

PR #40, opened 2017-11-16
Allow services to use kerberos KCM BZ(1512128)

lslebodn:
I didn't try to compile patch :-)

And it would be good to backport also to f27 and f26


wrabcak:
LGTM. 

Thanks.

PR #41, opened 2017-12-06
Allow qpid to map files (BZ1519948)

werwty:
Currently starting qpidd in f27 fails with
`AVC avc:  denied  { map } for  pid=5077 comm="qpidd" path="/var/lib/qpidd/.qpidd/qls/dat2/__db.001" dev="dm-0" ino=534085 scontext=system_u:system_r:qpidd_t:s0 tcontext=system_u:object_r:qpidd_var_lib_t:s0 tclass=file permissive=0`

audit2allow shows:
```
require {
    type qpidd_var_lib_t;
    type qpidd_t;
    class file map;
}

allow qpidd_t qpidd_var_lib_t:file map;
```

### DIFF
--- a/qpid.te
+++ b/qpid.te
@@ -48,6 +48,7 @@ manage_dirs_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 manage_files_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 manage_lnk_files_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 files_var_lib_filetrans(qpidd_t, qpidd_var_lib_t, { file dir lnk_file })
+allow qpidd_t qpidd_var_lib_t:file map;
 
 manage_dirs_pattern(qpidd_t, qpidd_var_run_t,  qpidd_var_run_t)
 manage_files_pattern(qpidd_t, qpidd_var_run_t,  qpidd_var_run_t)


qpid.te
```diff
@@ -29,6 +29,7 @@ files_pid_file(qpidd_var_run_t)
 # Local policy
 #
 
+allow qpidd_t qpidd_var_lib_t:file map;
```
wrabcak: Allow rule is correct, this will fix #BZ1519948 , but could you please add this rule to line 51 in this file? We're trying to have similar rules near. And this rule is connected to accessing qpidd_var_lib_t labels. 

Thanks. 
Lukas.

wrabcak (changes requested):
Fix is fine, just put it to the right line in the file.

werwty:
@wrabcak fixed!
Should I also open a PR to rawhide with this change?

qpid.te
```diff
@@ -48,6 +48,7 @@ manage_dirs_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 manage_files_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 manage_lnk_files_pattern(qpidd_t, qpidd_var_lib_t,  qpidd_var_lib_t)
 files_var_lib_filetrans(qpidd_t, qpidd_var_lib_t, { file dir lnk_file })
+allow qpidd_t qpidd_var_lib_t:file map;
```
wrabcak, line 51: Perfect!

wrabcak (approved):
LGTM.

wrabcak:
@werwty, 
Nope I'll back port it also to Rawhide. 

Thanks!

PR #45, opened 2018-01-09
Allow install_t to chat with systemd over D-Bus

jlebon:
An upcoming version of rpm-ostree will make use of systemd's D-Bus API.
Unfortunately, this is currently blocked by SELinux right now:

    subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 msg='avc:
    denied { send_msg } for msgtype=method_return dest=:1.23 spid=1
    tpid=1813 scontext=system_u:system_r:init_t:s0
    tcontext=unconfined_u:system_r:install_t:s0-s0:c0.c1023 tclass=dbus
    permissive=0

This patch adds a rule to allow rpm-ostree/anaconda to talk to systemd
over D-Bus.

### DIFF
--- a/anaconda.te
+++ b/anaconda.te
@@ -84,6 +84,7 @@ allow install_t self:capability2 mac_admin;
 
 systemd_dbus_chat_localed(install_t)
 systemd_dbus_chat_logind(install_t)
+init_dbus_chat(install_t)
 
 tunable_policy(`deny_ptrace',`',`
 	domain_ptrace_all_domains(install_t)


jlebon:
PR that will make use of this: https://github.com/projectatomic/rpm-ostree/pull/1147/
I successfully tested this patch in https://koji.fedoraproject.org/koji/taskinfo?taskID=24089792.

rhatdan:
LGTM

cgwalters:
Yep LGTM too!

rhatdan:
@wrabcak Can we get this into Fedora policy ASAP?

wrabcak:
Build will be later today. Sorry for late reply

cgwalters:
It doesn't look like there's a F27 build.  I'm a bit worried about the compat hazard here; I'd like to keep rpm-ostree releases monthly and if we have a hard dependency on things like this sepolicy change that gets harder.

I guess right now it'll just end up logging a denial...and only if someone has automatic updates enabled?  Eh...I guess we can live with that.

jlebon:
Hmm OTOH, it'd be nice if folks who wanted to experiment with the upcoming `check` mode didn't have to `setenforce 0`. (And https://github.com/projectatomic/rpm-ostree/issues/27 makes it tricky to cleanly "carry" a local module like we do in our tests). Let's say, if there's an selinux-policy bodhi update during the next rpm-ostree release, let's bind them together but otherwise we'll go ahead?

PR #47, opened 2018-01-27
Allow keepalived_t create and write a file under /tmp

masatake:
keepalived writes out its running statics to /tmp/keepalived.data when
it receives SIGUSR2. This change allows it.

Signed-off-by: Masatake YAMATO <yamato@redhat.com>

### DIFF
--- a/keepalived.te
+++ b/keepalived.te
@@ -18,6 +18,9 @@ files_pid_file(keepalived_var_run_t)
 type keepalived_unconfined_script_exec_t;
 application_executable_file(keepalived_unconfined_script_exec_t)
 
+type keepalived_tmp_t;
+files_tmp_file(keepalived_tmp_t)
+
 ########################################
 #
 # keepalived local policy
@@ -62,6 +65,9 @@ dev_read_urand(keepalived_t)
 modutils_domtrans_insmod(keepalived_t)
 
 logging_send_syslog_msg(keepalived_t)
+
+allow keepalived_t keepalived_tmp_t:file { create_file_perms write_file_perms };
+files_tmp_filetrans(keepalived_t, keepalived_tmp_t, file)
 
 optional_policy(`
     iptables_domtrans(keepalived_t)


keepalived.te
```diff
@@ -31,6 +34,7 @@ allow keepalived_t self:netlink_netfilter_socket create_socket_perms;
 allow keepalived_t self:netlink_route_socket nlmsg_write;
 allow keepalived_t self:packet_socket create_socket_perms;
 allow keepalived_t self:rawip_socket create_socket_perms;
+allow keepalived_t keepalived_tmp_t:file { create_file_perms write_file_perms };
```
wrabcak: This line should be removed and used manage_files_pattern macro like on line 39.

keepalived.te
```diff
@@ -63,6 +67,8 @@ modutils_domtrans_insmod(keepalived_t)
 
 logging_send_syslog_msg(keepalived_t)
 
+files_tmp_filetrans(keepalived_t, keepalived_tmp_t, file)
```
wrabcak: This is OK.

wrabcak (changes requested):

masatake:
Thank you for reviewing. 

keepalived just creates and writes to /tmp/keepalived.data. It never reads the file.
It overwrites the file if it exists. ( I tested the temporary file is created well with my patch.)
`manage_file_perms` gives more permissions to `keepalived_t`. Should I use `manage_files_pattern` macro anyway?

wrabcak:
Ok then, could move allow rule from line 37 to line 69? I'll merge it. 

Thanks,
Lukas

wrabcak:
Ok then, could move allow rule from line 37 to line 69? I'll merge it. 

Thanks,
Lukas

masatake:
Updated.

wrabcak:
LGTM. 
Thanks.

masatake:
Thank you.

PR #49, opened 2018-02-06
Allow certmonger_t domain to access /etc/pki/pki-tomcat BZ(1542600)

tiran:
Signed-off-by: Christian Heimes <cheimes@redhat.com>

See discussion on https://bugzilla.redhat.com/show_bug.cgi?id=1542600
certmonger requires DAC override because the NSSDB of Dogtag PKI is only accessible by user pkiuser and group pkiuser.

### DIFF
--- a/certmonger.te
+++ b/certmonger.te
@@ -32,7 +32,8 @@ files_tmp_file(certmonger_tmp_t)
 # Local policy
 #
 
-allow certmonger_t self:capability { chown  dac_read_search setgid setuid kill sys_nice };
+# certmonger requires DAC override to access /etc/pki/pki-tomcat/alias
+allow certmonger_t self:capability { chown dac_read_search dac_override setgid setuid kill sys_nice };
 dontaudit certmonger_t self:capability sys_tty_config;
 allow certmonger_t self:capability2 block_suspend;
 


lslebodn:
There is also a different way. certmonger is running as root but does not have to these directories with DAC. Therefore it requires such capability.

```
sh# rpm -q selinux-policy
selinux-policy-3.14.1-5.fc28.noarch

sh# ls -ld /etc/pki/pki-tomcat
drwxrwx---. 5 pkiuser pkiuser 4096 Feb  6 12:44 /etc/pki/pki-tomcat
```
Just user *pkiuser* or group *pkiuser* have access to such directory based on DAC.
So more secure approach will be to either run certmonger as pkiuser (which is created by dogtag)
or better to change group to root on related directories.

```
ls -ld /etc/pki/pki-tomcat
drwxrwx---. 5 pkiuser root 4096 Feb  6 12:44 /etc/pki/pki-tomcat
```

@wrabcak  do you agree?

tiran:
Please keep the discussion on BZ.

lslebodn:
As you wish

wrabcak:
@lslebodn 
I agree with you, but from BZ discussion I have no other option just to allow / merge it.

PR #50, opened 2018-02-09
Update pkcs policy to include pkccsslotd.service

dodys:
pkcsslotd.service was running, incorrectly, with default systemd label. Fixed it
by creating the pkcs_slotd_unit_file_t type and updating the file context.
This is also available on Refpolicy:
https://github.com/TresysTechnology/refpolicy-contrib/pull/67/commits/359430e4b9cac406b329fab8c68a574f80cd5ecf


wrabcak:
LGTM. 

Thanks.

PR #53, opened 2018-04-12
Allow git to mmap git_(sys|user)_content_t files BZ(1518027)

tmzullinger:
I tested this on f28 with `git-daemon`.  I tested clone, fetch, and ls-remote on system repositories (in the default /var/lib/git) and user repositories (in ~user/public_git).

I did not test with `git-http-backend` so I'm not sure if that requires anything further.

I'm also not sure whether the allow for `git_session_t` is needed or not.

PR #54, opened 2018-04-15
Allow callers of git_read_generic_sys_content_files() to map git_sys_content_t BZ(1518027)

tmzullinger:
Fix whitespace of read_lnk_files_pattern() call while here.

This follows up on #53 and should fix usage of git-http-backend (i.e. git smart http).  Sorry I didn't find this before submitting the initial pull request.  Hopefully this resolves the mmap AVC's for git.  I also see that lighttpd and nginx appear to be covered in the apache module.  I only tested with apache.


wrabcak:
LGTM 
THank!

PR #56, opened 2018-05-05
 Allow sssd_selinux_manager_t domain to chat with systemd 

lslebodn:
It would be good to backport these patches also to f26+


sssd.te
```diff
@@ -184,6 +184,9 @@ optional_policy(`
 # sssd SELinux manager local policy
 #
 
+# allow sssd_t to kill unresponsive selinux_child process
+allow sssd_t sssd_selinux_manager_t:process signal;
```
wrabcak: LGTM.

wrabcak (approved):

PR #57, opened 2018-05-08
Add tpm2-abrmd policy

martinezjavier:
The tpm2-abrmd is a user-space D-Bus based TPM2 Access Broker and Resource
Manager. Applications communicate with the tpm2-abrmd by sending commands
and receiving responses using a communication library (TCTI) that used to
create a pipe(2) to send the data to the daemon.

But this communication mechanism was changed to use unix sockets instead
and pass the file descriptors to the D-Bus daemon, so a SELinux policy is
needed to update the tpm2-abrmd to the latest version that use sockets.

This patch is based on the SELinux policy module that's in the tpm2-abrmd
upstream repository, with the changes suggested by Dominick Grift.

Suggested-by: Dominick Grift <dac.override@gmail.com>
Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/dbus.te
+++ b/dbus.te
@@ -236,6 +236,10 @@ optional_policy(`
     unconfined_server_domtrans(system_dbusd_t)
 ')
 
+optional_policy(`
+        tabrmd_rw_inherited_unix_stream_sockets(system_dbusd_t)
+')
+
 ########################################
 #
 # system_bus_type rules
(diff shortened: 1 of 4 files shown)


martinezjavier:
@wrabcak any comments on this?

wrabcak:
Hi @martinezjavier , 

Sorry for late reply. I would prefer to ship together with tpm2-abrmd package. This way brings several benefits mentioned here[1].

I saw your bug for new package review. Have it on my TODO. 

Closing this ticket.

[1] https://fedoraproject.org/wiki/SELinux/IndependentPolicy

PR #60, opened 2018-06-12
Add log file type to collectd and allow corresponding access

vmojzis:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1359061


wrabcak:
Merged. 
Thanks.

PR #61, opened 2018-06-22
Enable conntrackd to be NET_ADMIN

mscherer:
Since it do access netfilter and various others part of the
network stack, the capability is kinda required.

Without it, it fail with "can't open channel socket"


mscherer:
Related RHBZ https://bugzilla.redhat.com/show_bug.cgi?id=1594221

PR #62, opened 2018-08-09
mysql: Add same SELinux context for symlink as binary

kubco2:
We are unsure if in case of mysql we need to have same label for backward compatible symlink /usr/sbin/mysqld as target /usr/libexec/mysqld. In case it is advised to. This PR adds it, otherwise close it unmerged.
Thanks.


wrabcak:
LGTM.

PR #65, opened 2018-08-25
Add read interfaces for mysqld_log_t that was added in commit df832bf

georou:
Hi, adding in some read and dontaudit interfaces since last month's commit (df832bf) now labels MySQL's log files. With only write to mysqld_log_t currently present.

Side note, I'm not 100% sure on also having append. Since write is already present, might as well cover it?

Thank you.


wrabcak:
LGTM.
THanks!

PR #66, opened 2018-08-29
Conntrackd need to load kernel module to work

mscherer:
Otherwise, it fail with the following audit log:

    type=AVC msg=audit(1535552219.384:140): avc:  denied  { module_request } for  pid=806 comm="conntrackd" kmod="nfnetlink-subsys-1" scontext=system_u:system_r:conntrackd_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=system permissive=0


wrabcak:
LGTM.

THanks!

PR #72, opened 2018-09-27
Add nagios_send_cmd interface to give access to the cmd fifo

mscherer:
The cmd fifo is used by software such as nsca to trigger alert
in nagios.


wrabcak:
@mscherer, 

Why you created new Label for it? I whink we can use nagios_spool_t for it.

mscherer:
It did seems cleaner to have a specific label for that access, rather than mixing that with "fiddling with the spool directory" access. But I see why we also do not want to have a explosion of labels, so if you prefer that we use the existing one, I guess that's fine for me.

PR #74, opened 2018-12-07
Add corecmd_exec_shell interface

zpytela:
Signed-off-by: Zdenek Pytela <zdenek@pytela.net>

### DIFF
--- a/arpwatch.te
+++ b/arpwatch.te
@@ -58,6 +58,8 @@ kernel_read_kernel_sysctls(arpwatch_t)
 kernel_read_proc_symlinks(arpwatch_t)
 kernel_request_load_module(arpwatch_t)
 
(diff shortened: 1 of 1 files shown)


wrabcak (changes requested):
Thanks for patch.

I have few comments:
- Commit msg is misleading. From commit msg it looks like you adding interface "corecmd_exec_shell interface". 
- From commit msg there it's not clear which SELinux domain is improved. e.g: " Allow domains labeled as arpwatch_t to exec shell"
- Could you please also add referecne github issue/bugzilla ticket (if possible) in commit msg? I need to know to which stable fedora I should backport it. 

Comment in PR what versions of Fedora are affected is appreciated. 

Otherwise patch is good.

wrabcak:
Closing because author created new pull request #75  with same fix.

PR #75, opened 2018-12-10
Allow arpwatch_t domains to execute shell BZ(1644568)

zpytela:
Sending e-mails is a part of core functionality of arpwatch.
If esmtp package is installed to provide the sendmail capability,
the permission to execute shell_exec_t for arpwatch_t is required
as esmtp actually provides a shell wrapper.

The arpwatch service runs as an unprivileged user, so granting
this permission should not be considered as too excessive.

Fedoras 29 and 28 are affected.

Signed-off-by: Zdenek Pytela <zdenek@pytela.net>

### DIFF
--- a/arpwatch.te
+++ b/arpwatch.te
@@ -58,6 +58,8 @@ kernel_read_kernel_sysctls(arpwatch_t)
 kernel_read_proc_symlinks(arpwatch_t)
 kernel_request_load_module(arpwatch_t)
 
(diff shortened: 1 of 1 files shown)


wrabcak:
LGTM. 
Thanks.

wrabcak (approved):

wrabcak:
Back ported to F29 and F28

PR #79, opened 2018-12-19
This is a proposed fix for https://jira.mongodb.org/browse/SERVER-38704

mattlord:
In MongoDB 3.4.16, 3.6.6, 4.0.0 and later, mongod tries to read netstat info from /proc/net and store it in its diagnostic system (FTDC). See: https://jira.mongodb.org/browse/SERVER-31400

This means that we need to adjust the policy so that the mongod process is allowed to open and read /proc/net/netstat, which also typically has symlinks (e.g. /proc/net/snmp).

### DIFF
--- a/mongodb.te
+++ b/mongodb.te
@@ -60,6 +60,15 @@ manage_files_pattern(mongod_t, mongod_tmp_t, mongod_tmp_t)
 manage_sock_files_pattern(mongod_t, mongod_tmp_t, mongod_tmp_t)
 files_tmp_filetrans(mongod_t, mongod_tmp_t, { file dir sock_file })
 
+## In MongoDB 3.4.16, 3.6.6, 4.0.0 and later, mongod tries to read netstat
+## info from /proc/net and store it in its diagnostic system (FTDC).
+## See: https://jira.mongodb.org/browse/SERVER-31400
+## This means that we need to adjust the policy so that the mongod
+## process is allowed to open and read /proc/net/netstat, which
+## also typically has symlinks (e.g. /proc/net/snmp).
+kernel_list_proc(mongod_t)
+kernel_read_proc_symlinks(mongod_t)
+
 kernel_read_system_state(mongod_t)
 kernel_read_network_state(mongod_t)
 kernel_read_vm_sysctls(mongod_t)


mongodb.te
```diff
@@ -60,6 +60,16 @@ manage_files_pattern(mongod_t, mongod_tmp_t, mongod_tmp_t)
 manage_sock_files_pattern(mongod_t, mongod_tmp_t, mongod_tmp_t)
 files_tmp_filetrans(mongod_t, mongod_tmp_t, { file dir sock_file })
 
+## In MongoDB 3.4.16, 3.6.6, 4.0.0 and later, mongod tries to read netstat
+## info from /proc/net and store it in its diagnostic system (FTDC).
+## See: https://jira.mongodb.org/browse/SERVER-31400
+## This means that we need to adjust the policy so that the mongod
+## process is allowed to open and read /proc/net/netstat, which
+## also typically has symlinks (e.g. /proc/net/snmp).
+allow mongod_t proc_net_t:file { open read };
```
wrabcak: You can use following interface "kernel_read_network_state" instead of using raw allow rule.

wrabcak (changes requested):
Please use mentioned interface instead of raw allow rule and I'll merge this. 

Thanks for you help!
Lukas

mongodb.te
mattlord: Hi @wrabcak,

It already has kernel_read_network_state, so I'm assuming you mean that these two new options are sufficient additions to that which then eliminates the need for the explicit allow:
`kernel_list_proc(mongod_t)`
`kernel_read_proc_symlinks(mongod_t)`

I'll adjust the PR accordingly for now and can correct again if I'm misunderstanding anything. 

Thank you for the help!

mattlord: OK, that's done. I'll leave the conversation open in case I misunderstood the request. Thanks again!

wrabcak:
Hi Mattlord, 

Could you squash these commits to 1 commit? Then, I'll merge it.

Thanks,
Lukas.

mattlord:
Hi Lukas,

I squashed my two commits and cleaned things up so that it should be an easy 1 commit merge. 

Thank you again for all of your help and patience! I appreciate it.  

Best Regards,

Matt

wrabcak (approved):

wrabcak:
Hi Mattlord, 

Thank you for help on SELinux policies. :) Backporting PR to Fedora 29 and Fedora 28.

PR #77, opened 2018-12-18
Add file transition for /var/lib/arpwatch/.esmtp_queue

zpytela:
Add file transition so that system_mail_t can in the /var/lib/arpwatch directory
with arpwatch_data_t context create .esmtp_queue directory with mail_home_rw_t context.
It is required for the arpwatch service being able to send emails.

Signed-off-by: Zdenek Pytela <zpytela@redhat.com>

PR #81, opened 2019-01-09
Allow arpwatch send e-mail notifications BZ(1657327)

zpytela:
Label /var/lib/arpwatch/.esmtp_queue as mail_home_rw_t and add file
transition so that system_mail_t can in the /var/lib/arpwatch directory
with arpwatch_data_t context create .esmtp_queue directory with
mail_home_rw_t context.

It is required for the arpwatch service being able to send email notifications
about changes in its database. If esmtp package is installed to provide the sendmail
rpm-capability, the ~arpwatch/.esmtp_queue directory has to have the right label
so that the daemon is able to send e-mails.

Fedoras 29 and 28 are affected.

### DIFF
--- a/arpwatch.if
+++ b/arpwatch.if
@@ -217,7 +217,7 @@ interface(`arpwatch_admin',`
 ##	</summary>
 ## </param
 #
-interface(`arpwatch_home_dir_filetrans',`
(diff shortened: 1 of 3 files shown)


zpytela:
Closing and following-up in the original request #77.

PR #85, opened 2019-02-06
Fix CI for use on forks

WOnder93:
Generally, I like to enable Travis CI on my forks so I know if I break
the build even before I open a PR. Howver, this repo's .travis.yml
uses a hard-coded path to the home directory, which depends on the name
of the repository owner.

Fix this to use $PWD instead, so that CI works also on forks.

Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
     - docker pull fedora:rawhide
 
 script:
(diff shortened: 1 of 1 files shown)


wrabcak:
LGTM. 
Thanks.

PR #88, opened 2019-02-15
Red Hat Bug 1627075 - Create a SELinux boolean to disable cron-logrotate transition

ngutlhtej:


ghost:
Ironically this adds a hard dependency on cron for logrotate.. Whereas without this commit you can, in theory at least, simply remove or disable the crond module (and better yet simply de-install cronie) if you want to use systemd timers instead ...

ghost:
... but aside from that. The issue described in 1627075 is not tied to logrotate. The various scripts in /etc/cron\\..* would ideally have a type that can be used by cronie to enter the system cronjob domain. Alternatively one can label /etc/cron.daily/logrotate type logrotate_exec_t, but that would not solve the issue for other scripts in those locations

ngutlhtej:
> Ironically this adds a hard dependency on cron for logrotate.. Whereas without this commit you can, in theory at least, simply remove or disable the crond module (and better yet simply de-install cronie) if you want to use systemd timers instead ...

How does this add a hard dependency? It was already there.
We want to disable TRANSITION from system_cronjob_t to logrotate_t because that type has no execstack permission, which is required for the IBM executable dsmc to work en SELinux-enabled RHEL. We have already reached system_cronjob_t and want to keep it this way, not transition to logrotate_t.

Whether we actually use cron or systemd timers is irrelevant as the associated SELinux types are essentially interchangeable.

ghost:
Benjamin Lefoul <notifications@github.com> writes:

>  Ironically this adds a hard dependency on cron for logrotate.. Whereas without this commit you can, in theory at least, simply remove or
>  disable the crond module (and better yet simply de-install cronie) if you want to use systemd timers instead ...
>
> How does this add a hard dependency? It was already there.
> We want to disable TRANSITION from system_cronjob_t to logrotate_t because that type has no execstack permission, which is required for the IBM
> executable dsmc to work en SELinux-enabled RHEL. We have already reached system_cronjob_t and want to keep it this way, not transition to
> logrotate_t.

The commit removes the 'optional_policy(`'?

```execstack

    As the name suggests, this error is raised if a program tries to make its stack (or parts thereof) executable with an mprotect call. This should never, ever be necessary. Stack memory is not executable on most OSes these days and this won't change. Executable stack memory is one of the biggest security problems. An execstack error might in fact be most likely raised by malicious code.

    See my overview of security features in Fedora and RHEL for more information, specifically appendix A. It explains how to avoid executable stacks.```

https://akkadia.org/drepper/selinux-mem.html

Note the "This should never, ever be necessary"

>
> Whether we actually use cron or systemd timers is irrelevant as the associated SELinux types are essentially interchangeable.
>
> —
> Reply to this email directly, view it on GitHub, or mute the thread.
>

-- 
Dominick Grift

ngutlhtej:
> The commit removes the 'optional_policy(`'? 

...how does that had a hard dependency? This is independent from RPM spec.

> ```execstack As the name suggests, this error is raised if a program tries to make its stack (or parts thereof) executable with an mprotect call. This should never, ever be necessary. Stack memory is not executable on most OSes these days and this won't change. Executable stack memory is one of the biggest security problems. An execstack error might in fact be most likely raised by malicious code. See my overview of security features in Fedora and RHEL for more information, specifically appendix A. It explains how to avoid executable stacks.``` https://akkadia.org/drepper/selinux-mem.html Note the "This should never, ever be necessary"

I... know? I mentioned it in the bug report months ago. You did say you read it...


Please, if you are not from the Red Hat ticket 1627075 or the IBM ticket TS001533525, just comment if you have something useful to say.

zpytela:
Closing this PR along with the resolution of RHEL bz https://bugzilla.redhat.com/show_bug.cgi?id=1672199

PR #89, opened 2019-02-19
[WIP] Add mozilla_plugin_use_samba boolean BZ(1642423)

zpytela:
Signed-off-by: Zdenek Pytela <zpytela@redhat.com>

PR #90, opened 2019-02-20
Fb samba manage share if

zpytela:
Add samba_manage_share_files interface to help with bz 1642423 and
https://github.com/fedora-selinux/selinux-policy-contrib/pull/89

### DIFF
--- a/samba.if
+++ b/samba.if
@@ -440,6 +440,26 @@ interface(`samba_read_share_files',`
 
 ########################################
 ## <summary>
+##	Allow the specified domain to manage samba's shares
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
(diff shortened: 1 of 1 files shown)


zpytela:
The package with this commit builds, installs, and makes #89 applicable. Please review them together.

ghost:
I think this is a misconfiguration.
The reporter probably should have not labeled his home directory type samba_share_t but should probably instead have set the `samba_enable_home_dirs` boolean.

Other than that:
1. you have a rule that allows the caller to get attributes of samba_share_t filesystems and there is no mention of this in the bugreport AFAICT, secondly there is no samba_share_t filesystem (its cifs_t)
2. your interface name implies that the interface applies to "manage samba share files" but the interface also allow for managing "samba share dirs"

PR #92, opened 2019-03-19
Add support for smart card authentication in cockpit BZ(1690444)

zpytela:

### DIFF
--- a/cockpit.fc
+++ b/cockpit.fc
@@ -5,6 +5,7 @@
 
 /usr/libexec/cockpit-ws		--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
 
+/usr/libexec/cockpit-cert-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
 /usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
 /usr/libexec/cockpit-ssh	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
 
--- a/cockpit.te
+++ b/cockpit.te
@@ -96,6 +96,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+    miscfiles_dontaudit_map_generic_certs(cockpit_ws_t)
+')
+
+optional_policy(`
     systemd_exec_systemctl(cockpit_ws_t)
 ')
 
@@ -134,6 +138,10 @@ corenet_tcp_connect_ssh_port(cockpit_session_t)
 # cockpit-session can execute cockpit-agent as the user
 userdom_spec_domtrans_all_users(cockpit_session_t)
 usermanage_read_crack_db(cockpit_session_t)
+
+optional_policy(`
+    sssd_dbus_chat(cockpit_session_t)
+')
 
 optional_policy(`
     userdom_signal_all_users(cockpit_session_t)


zpytela:
Scratchbuild finished and installed. Running
rpm -qa "selinux-policy*"; matchpathcon /usr/libexec/cockpit-cert-session; sesearch -A -s cockpit_session_t -t sssd_t -c dbus -p send_msg; sesearch -A -s sssd_t -t cockpit_session_t -c dbus -p send_msg

before:
selinux-policy-3.14.2-51.fc29.noarch
selinux-policy-targeted-3.14.2-51.fc29.noarch
selinux-policy-devel-3.14.2-51.fc29.noarch
selinux-policy-doc-3.14.2-51.fc29.noarch
/usr/libexec/cockpit-cert-session       system_u:object_r:bin_t:s0

after:
selinux-policy-3.14.4-3.1690444.200.fc31.noarch
selinux-policy-devel-3.14.4-3.1690444.200.fc31.noarch
selinux-policy-targeted-3.14.4-3.1690444.200.fc31.noarch
selinux-policy-doc-3.14.4-3.1690444.200.fc31.noarch
/usr/libexec/cockpit-cert-session       system_u:object_r:cockpit_session_exec_t:s0
allow cockpit_session_t sssd_t:dbus send_msg;
allow sssd_t cockpit_session_t:dbus send_msg;

Please review.

zpytela:
F30 should be the target release.

martinpitt:
Hang on, on Fedora 29 (not on RHEL 8 or anywhere else) I get these two violations. They don't break the operation, but still cause unexpected log errors:
```
audit: type=1400 audit(1553006049.704:621): avc:  denied  { map } for  pid=8208 comm="pool" path="/etc/pki/ca-trust/source/README" dev="dm-0" ino=8538086 scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:object_r:cert_t:s0 tclass=file permissive=0
audit: type=1400 audit(1553006049.710:622): avc:  denied  { map } for  pid=8208 comm="pool" path="/etc/pki/ca-trust/source/ipa.p11-kit" dev="dm-0" ino=9401424 scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:object_r:cert_t:s0 tclass=file permissive=0
```
I'm not sure what "pool" is (there is no such binary), but I'm fairly sure that this is cockpit's web server trying to validate the TLS certificate, through glib-networking → gnutls.

I'll play around with that a bit and see how to extend the policy to allow this.

Thanks!

martinpitt:
Indeed, allowing `cockpit_ws_t` to read `cert_t`  files helps. This is probably best expressed as a diff to my original "local.te" file:

```diff
-require { type cockpit_session_t; type sssd_t; class dbus send_msg; }
+require { type cockpit_session_t; type sssd_t; type cockpit_ws_t; type cert_t; class dbus send_msg; class file { read map }; }
 allow cockpit_session_t sssd_t:dbus send_msg;
 allow sssd_t cockpit_session_t:dbus send_msg;
+allow cockpit_ws_t cert_t:file { read map };
```

zpytela:
@martinpitt 
Are these permissions required for supporting the smart card auth feature?
Are these permissions necessary to be granted to have it working?

As a sidenote, it is quite hard to believe some piece of software needs to map a README file.

martinpitt:
@zpytela : No, as I wrote these don't break the operation. cockpit-ws does not actually need to look up the system certificate chain to validate trust anchors, it just checks whether the certificate is valid in itself (e. g., not expired). It leaves the actual yay/nay decision to sssd. So these last rules are not required per se, it's just a bit ugly to get SELinux violations in the journal. So if the policy can be written to explicitly deny  these, so that the violation messages don't  happen, that  works as well.

GnuTLS certainly doesn't look for the README file in particular, I suppose it just scans the entire /etc/pki/ca-trust/ directory.

zpytela:
@martinpitt: In addition to allow rules there can be also dontaudit rules in the policy which serve exactly this purpose - the permission is not granted, but the attempt is not audited. I just wanted to doublecheck it is the case as dontaudit rules a bit more troublesome when troubleshooting.

I would also like to have confirmed it is still the same task or goal to pair it properly - in the second part there is a different domain cockpit_ws_t whereas in the first one it was cockpit_session_t.

martinpitt:
@zpytela : Right, so in that case `dontaudit` seems appropriate. That means "explicitly deny cockpit-ws to read /etc/pki", which is fine.

The different types are deliberate. It's the unprivileged `/usr/libexec/cockpit-ws` (`cockpit_ws_t`) web server which does the TLS termination with GnuTLS, which then calls the suid root `/usr/libexec/cockpit-session` (`cockpit_session_t`) to do the actual PAM/sssd query/session stuff.

zpytela:
@martinpitt: I updated the pull request and created another one for the base policy. Can you now have cockpit running for some time to test if some other denials appear with your local policy in place or would you like to wait for the policy update?

martinpitt:
I grabbed the current selinux-policy fedpkg (f29 branch), updated it to your repo/branch, and tried to build it. It complained about missing ejabberd module, that was simple enough to drop from modules-targeted-contrib.conf. But now it fails with
```
/usr/bin/checkmodule -M -U allow base.conf -o tmp/base.mod
policy/modules/kernel/domain.te:609:ERROR 'syntax error' at token 'vmtools_unconfined_dbus_chat' on line 173013:
#line 609
    vmtools_unconfined_dbus_chat(domain)
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
```
kernel/domain.te is not even defined in contrib, but in the main policy, which didn't change.

So I just applied the patch from this PR and from https://github.com/fedora-selinux/selinux-policy/pull/252/files to selinux-policy-contrib-adc457b.tar.gz, mockbuilt, and installed it into my test VM instead of the manual policy hack. All tests work, and there are no unexpected SELinux violations.

As I spent quite a lot of time on building this package (I fell into a lot of newbie traps admittedly), I conserved this as a [scatch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=33686999), so that this is easier for my future self and others to test.

Thank you!

wrabcak:
LGTM. 
Thanks.

PR #94, opened 2019-03-21
Add tcpd_wrapped_domain for telnetd BZ(1676940)

zpytela:
Signed-off-by: Zdenek Pytela <zpytela@redhat.com>

PR #95, opened 2019-03-22
Allow tcpd bind to services ports BZ(1676940)

zpytela:
tcpd from tcp_wrappers is a network services superserver
working similar to inet.
tcpd_t needs to bind to ports for those services where a transition
from tcpd_t exists in the policy.

This patch is needed to complete BZ(1676940).


zpytela:
Requesting review of this PR. See
https://github.com/fedora-selinux/selinux-policy-contrib/pull/94

for further details.

wrabcak:
LGTM.  
Thanks.

PR #100, opened 2019-04-01
Allow fontconfig file transition for xguest_u user

zpytela:


wrabcak:
LGTM. Thanks.

But we should investigate if we need to allow this for more user domains.

PR #101, opened 2019-04-02
Fix varnisncsa typo

dridi:
This may lead varnishncsa.service failing in enforcing mode.

### DIFF
--- a/varnishd.fc
+++ b/varnishd.fc
@@ -5,7 +5,7 @@
 /etc/varnish(/.*)?	gen_context(system_u:object_r:varnishd_etc_t,s0)
 
 /usr/bin/varnishlog	--	gen_context(system_u:object_r:varnishlog_exec_t,s0)
-/usr/bin/varnisncsa	--	gen_context(system_u:object_r:varnishlog_exec_t,s0)
+/usr/bin/varnishncsa	--	gen_context(system_u:object_r:varnishlog_exec_t,s0)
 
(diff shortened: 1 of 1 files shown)


wrabcak:
LGTM. 

Thanks,
Lukas

dridi:
That was quick, thanks. I checked the policy and I don't think varnishncsa.service will fail with this change.

The problems I was foreseeing were fixed last July and now that `/usr/bin/varnishncsa` is properly labeled it should be able to find and map `varnishd`'s shared memory.

dridi:
For reference:

- https://github.com/fedora-selinux/selinux-policy-contrib/commit/5f8cfba0a6d4a7f67caeb590d3841eb5700ea402
- https://github.com/fedora-selinux/selinux-policy-contrib/commit/b0969338a3c75154424169b9df6231f67db37c6a

dridi:
Actually, I think we need this:

```
allow varnishlog_t varnishd_var_lib_t:dir read;
```

wrabcak:
Can I expect PR? :)

dridi:
Well, I need this PR to land in rawhide so I can verify this, and if I'm right I will send a PR.

How can I be informed? (fellow Fedora contributor here=)

dridi:
Actually I can remove the varnishd module and try my own policy to check this. I will try to allocate some time to bring up a rawhide environment and give this a try.

PR #105, opened 2019-05-02
Allow systemd to execute sa-update in spamd_update_t domain BZ(1705331)

zpytela:
Replace application_domain with init_system_domain for spamd_update_exec_t.
The /usr/share/spamassassin/sa-update.cron file is now labeled with
spamd_update_exec_t since commit db62634 and it is called off a timer unit file.


wrabcak:
LGTM

PR #106, opened 2019-05-02
Add fcontext for apachectl util to fix missing output

uhliarik:
when executed "httpd -t" from this script.


wrabcak:
LGTM. Thanks

PR #107, opened 2019-05-03
Allow pcp_pmie_t domain to use sys_ptrace usernamespace cap

zpytela:


zpytela:
Rawhide, Fedora 30 and 29 are affected.

lberk:
LGTM

We've been carrying around a patch like this in upstream PCP for some time now, however it's conditional due to not being available in every distro.  Having the rule upstream will definitely simplify PCP's upstream policy a bit, thanks!

wrabcak:
@lberk  Thanks for review! :) 

@zpytela, 
PR branch conflicts with the master, could you please take a look and fix it? 

Thanks,
Lukas.

zpytela:
Hope it is in a better shape now, @wrabcak

wrabcak:
LGTM. THanks.

PR #108, opened 2019-05-07
Add tlp_filetrans_named_content interface BZ(1705246)

zpytela:
Interface required for another pr which follows.


zpytela:
File transition is not needed if there is a process transition, closing.

PR #110, opened 2019-05-16
Grant varnishlog_t access to varnishd_etc_t

dridi:
The varnishncsa program has a -f \<file\> option that reads the NCSA
format string from a file to work around shell expansion pitfalls of
its -F \<format\> counterpart. For that the logical location would be
/etc/varnish where other varnish configuration files live.

It is fine to also grant varnishlog access to this location. Even
though there is currently no use case for it today, there are plans
that would require this access too.

This applies to all active Fedora branches today.

### DIFF
--- a/varnishd.te
+++ b/varnishd.te
@@ -144,3 +144,4 @@ read_files_pattern(varnishlog_t, varnishd_var_lib_t, varnishd_var_lib_t)
 allow varnishlog_t varnishd_var_lib_t:file map;
 
 files_search_var_lib(varnishlog_t)
+varnishd_read_config(varnishlog_t)


ghost (commented):
Probably best to use `varnishd_read_config(varnishlog_t)` because your rules are incomplete, also varnish config lives in a `/etc/varnish` which is labeled `varnishd_etc_t`. Your rules do not allow it to traverse that.

dridi:
Indeed this is probably my mistake.

The original policy I wrote to get it working was:

```
allow varnishlog_t varnishd_etc_t:dir search;
allow varnishlog_t varnishd_etc_t:file { open read };
```

That was the quick-n-dirty change to the policy I came up with on the spot and thought that I should rather submit something similar to what was already present. My bad.

I will give a try to `varnishd_read_config()` and update the pull request accordingly. We could possibly simplify the policy further with `varnishd_read_lib_files()` too? I will also try that before coming back to you.

Thanks for the prompt response!

In the mean time I actually started working on a feature that would require `varnishlog` access to `/etc/varnish` too: https://github.com/varnishcache/varnish-cache/pull/3001.

dridi:
I have successfully tested the following diff:

```diff
diff varnishd.if varnishd.if
--- varnishd.if
+++ varnishd.if
@@ -75,7 +75,7 @@ interface(`varnishd_read_lib_files',`
        ')
 
        files_search_var_lib($1)
-       read_files_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
+       mmap_read_files_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
 ')
 
 #######################################
diff varnishd.te varnishd.te
--- varnishd.te
+++ varnishd.te
@@ -169,15 +169,9 @@ create_files_pattern(varnishlog_t, varnishlog_log_t, varnishlog_log_t)
 setattr_files_pattern(varnishlog_t, varnishlog_log_t, varnishlog_log_t)
 logging_log_filetrans(varnishlog_t, varnishlog_log_t, { file dir })
 
-read_files_pattern(varnishlog_t, varnishd_var_lib_t, varnishd_var_lib_t)
-
-list_dirs_pattern(varnishlog_t, varnishd_var_lib_t, varnishd_var_lib_t)
-files_search_var_lib(varnishlog_t)
-allow varnishlog_t varnishd_var_lib_t:dir read;
-allow varnishlog_t varnishd_var_lib_t:file map;
+varnishd_read_config(varnishlog_t);
+varnishd_read_lib_files(varnishlog_t)
```

The rationale for `mmap_read_files_pattern()` is that most files that live in `/var/lib/varnish` are meant to be `mmap`'ed and this is done transparently by `libvarnishapi.so`, which is used by `varnishlog` and `varnishncsa`. So any consumer of `varnishd_read_lib_files()` ultimately needs to add a `map` permission anyway.

I will update the pull request as soon as possible but do not hesitate to share your review again.

ghost:
On Mon, May 20, 2019 at 03:19:02AM -0700, Dridi Boukelmoune wrote:
> I have successfully tested the following diff:
> 
> ```diff
> diff varnishd.if varnishd.if
> --- varnishd.if
> +++ varnishd.if
> @@ -75,7 +75,7 @@ interface(`varnishd_read_lib_files',`
>         ')
>  
>         files_search_var_lib($1)
> -       read_files_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
> +       mmap_read_files_pattern($1, varnishd_var_lib_t, varnishd_var_lib_t)
>  ')
>  
>  #######################################
> diff varnishd.te varnishd.te
> --- varnishd.te
> +++ varnishd.te
> @@ -169,15 +169,9 @@ create_files_pattern(varnishlog_t, varnishlog_log_t, varnishlog_log_t)
>  setattr_files_pattern(varnishlog_t, varnishlog_log_t, varnishlog_log_t)
>  logging_log_filetrans(varnishlog_t, varnishlog_log_t, { file dir })
>  
> -read_files_pattern(varnishlog_t, varnishd_var_lib_t, varnishd_var_lib_t)
> -
> -list_dirs_pattern(varnishlog_t, varnishd_var_lib_t, varnishd_var_lib_t)
> -files_search_var_lib(varnishlog_t)
> -allow varnishlog_t varnishd_var_lib_t:dir read;
> -allow varnishlog_t varnishd_var_lib_t:file map;
> +varnishd_read_config(varnishlog_t);

please remove the ;

> +varnishd_read_lib_files(varnishlog_t)

With this you are removing the permission that allows varnishlog_t to list the directory. Not sure if this permission was needed in the first place.

Another *personal* issue i have (the maintainer might disagree) is that you add the map permission to the existing varnishd_read_var_lib_files
I would have created a new `varnishd_map_read_var_lib_files()` instead.

```
interface(`varnishd_map_read_lib_files',`
	varnishd_read_lib_files($1)
	allow $1 varnishd_var_lib_t:file map;
')
```

But again this is personal preference (i dont like "breaking"/"changing" existing interfaces) but I am just an independent individual reviewing.

> ```
> 
> The rationale for `mmap_read_files_pattern()` is that most files that live in `/var/lib/varnish` are meant to be `mmap`'ed and this is done transparently by `libvarnishapi.so`, which is used by `varnishlog` and `varnishncsa`. So any consumer of `varnishd_read_lib_files()` ultimately needs to add a `map` permission anyway.
> 
> I will update the pull request as soon as possible but do not hesitate to share your review again.
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/fedora-selinux/selinux-policy-contrib/pull/110#issuecomment-493926534

-- 
Dominick Grift

dridi:
> With this you are removing the permission that allows varnishlog_t to list the directory. Not sure if this permission was needed in the first place.

I would rather add this permission to the `varnishd_read_lib_files()` interface in this case. I'm not sure Varnish needs to list the contents of a directory, since `libvarnishapi.so` starts from a specific file from the directory and then finds references to other files from its (`mmap`'ed) contents.

> Another *personal* issue i have (the maintainer might disagree) is that you add the map permission to the existing varnishd_read_var_lib_files I would have created a new `varnishd_map_read_var_lib_files()` instead.

To be picky I would call it `varnishd_mmap_read_var_lib_files()` (with an extra `m`) to match the underlying construct. But in its current form `varnishd_read_var_lib_files()` is pointless by itself since the contents of `/var/lib/varnish` are ultimately `mmap`'ed so introducing a new interface would bring little value. I would otherwise recommend to deprecate the old one in favor of the new one.

> But again this is personal preference (i dont like "breaking"/"changing" existing interfaces) but I am just an independent individual reviewing.

And I'm neither an SELinux expert, nor familiar with how the policy is maintained. So I will naturally follow the decisions from the review, and I tried to explain what "reading" `varnishd_var_lib_t` implies.

ghost:
On Mon, May 20, 2019 at 04:37:49AM -0700, Dridi Boukelmoune wrote:
> > With this you are removing the permission that allows varnishlog_t to list the directory. Not sure if this permission was needed in the first place.
> 
> I would rather add this permission to the `varnishd_read_lib_files()` interface in this case. I'm not sure Varnish needs to list the contents of a directory, since `libvarnishapi.so` starts from a specific file from the directory and then finds references to other files from its (`mmap`'ed) contents.
> 
> > Another *personal* issue i have (the maintainer might disagree) is that you add the map permission to the existing varnishd_read_var_lib_files I would have created a new `varnishd_map_read_var_lib_files()` instead.
> 
> To be picky I would call it `varnishd_mmap_read_var_lib_files()` (with an extra `m`) to match the underlying construct. But in its current form `varnishd_read_var_lib_files()` is pointless by itself since the contents of `/var/lib/varnish` are ultimately `mmap`'ed so introducing a new interface would bring little value. I would otherwise recommend to deprecate the old one in favor of the new one.
> 
> > But again this is personal preference (i dont like "breaking"/"changing" existing interfaces) but I am just an independent individual reviewing.
> 
> And I'm neither an SELinux expert, nor familiar with how the policy is maintained. So I will naturally follow the decisions from the review, and I tried to explain what "reading" `varnishd_var_lib_t` implies.

Fair enough, I will leave "var lib" aspect to others to decide.

The config aspect is addressed AFAIC (the issue was not controversial or subjective)

> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/fedora-selinux/selinux-policy-contrib/pull/110#issuecomment-493950087

-- 
Dominick Grift

dridi:
Thank you for your review, as promised I will eventually update the pull request. I can open a separate one for the var lib question.

dridi:
Pull request updated, taking only care of the original issue at hand.

I will open a separate pull request for the `mmap` question, or push an additional commit to this pull request, depending on what you prefer.

wrabcak:
Thanks for help with review @doverride .

LGTM. 
Thanks!

dridi:
I'll take that as a hint that you prefer a separate pull request for the other issue discussed here. Probably early next week unless I manage to salvage some time before.

PR #112, opened 2019-05-27
Dontaudit net_admin capability for confined users dbusd type

zpytela:
The rule has been moved from sysadm.te to dbus.if/dbus_role_template
to support the suggested resolution of BZ(1703241)


wrabcak:
LGTM.

PR #114, opened 2019-06-14
cockpit: Support split-out TLS proxy

martinpitt:
Soon, the TLS termination and client-side certificate handling will be
split out into a new cockpit-tls process. This communicates with
cockpit-ws via a Unix domain socket.

------

Without this, I'm currently getting
```
denied  { connectto } for  pid=6668 comm="cockpit-tls" path="/run/cockpit/tls/ws.8.sock" scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:system_r:cockpit_ws_t:s0 tclass=unix_stream_socket permissive=0

denied  { execute_no_trans } for  pid=6759 comm="cockpit-tls" path="/usr/libexec/cockpit-ws" dev="dm-0" ino=1505205 scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:object_r:cockpit_ws_exec_t:s0 tclass=file permissive=0
```

I was testing with this local policy adjustment:
```sh
semanage fcontext -a /usr/libexec/cockpit-tls -t cockpit_ws_exec_t
restorecon /usr/libexec/cockpit-tls

cat <<EOF > /tmp/local.te
module local 1.0;
require {
    type cockpit_ws_t;
    type cockpit_ws_exec_t;
    class unix_stream_socket { create_stream_socket_perms connectto };
    class file { execute_no_trans};
}

allow cockpit_ws_t cockpit_ws_t:unix_stream_socket { create_stream_socket_perms connectto };
allow cockpit_ws_t cockpit_ws_exec_t:file { execute_no_trans };
EOF
checkmodule -M -m -o /tmp/local.mod /tmp/local.te
semodule_package -o /tmp/local.pp -m /tmp/local.mod
semodule -i /tmp/local.pp
```

I haven't yet directly tested this PR (this is rather cumbersome, and I can't do it right now -- I'll be able to on Monday). Does this look reasonable?

Thanks!

### DIFF
--- a/cockpit.fc
+++ b/cockpit.fc
@@ -4,6 +4,7 @@
 /etc/systemd/system/cockpit.*	--	gen_context(system_u:object_r:cockpit_unit_file_t,s0)
 
 /usr/libexec/cockpit-ws		--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-tls	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
 
 /usr/libexec/cockpit-cert-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
 /usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
--- a/cockpit.te
+++ b/cockpit.te
@@ -40,6 +40,9 @@ allow cockpit_ws_t self:tcp_socket create_stream_socket_perms;
 
 kernel_read_system_state(cockpit_ws_t)
 
+# cockpit-tls can execute cockpit-ws
+can_exec(cockpit_ws_t,cockpit_ws_exec_t)
+
 # cockpit-ws can execute cockpit-session
 can_exec(cockpit_ws_t,cockpit_session_exec_t)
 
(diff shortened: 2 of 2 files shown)


WOnder93:
Wouldn't it be better to create a new separate domain for the TLS process (say, `cockpit_tls_t`) to have a more granular policy?

martinpitt:
This is still under heavy development, and it's not clear yet how this will eventually shape out. We can't completely stop cockpit-ws from doing TLS by itself, as it may still be called directly from e. g. Satellite or ManagedIQ, custom Apache configs or other reverse proxies, so it would mean to duplicate most of the policy for cockpit-tls. At that time this seems too unwieldy to me.

Once this all landed and settled down, we may reinvestigate tightening the policy, of course. This is not actually be that far out, as eventually (soon) we want to run cockpit-tls as a separate system user `cockpit-tls` instead of `cockpit-ws`, then (I think) we have to adjust the policy again anyway, and can consider a tighter split.

WOnder93:
OK, that makes sense, thanks!

wrabcak:
Hi Guys, 
Thanks for explaining, it make sense and looks good to me. Feel free to contact us for future splitting the policy. 
Thanks,
Lukas.

PR #117, opened 2019-06-21
Change condor_domain declaration in condor_systemctl

zpytela:
Change condor_domain declaration from type to attribute
in condor_systemctl() interface


zpytela:
The package with the fix builds and the custom module using condor_admin(confined_admin_t,confined_admin_r) does not trigger the error any longer. Please review.

wrabcak:
LGTM, 
Thanks. 

Merging also to F30 and F29.

PR #118, opened 2019-06-21
Replace "-" by "_" in types names

zpytela:
Replace "-" by "_" in all names of the speech-dispatcher.*_t types
as hyphen is not understood by checkmodule when used in a custom policy


zpytela:
After updating the selinux policy packages it is possible to use the speech_dispatcher_admin interface in a custom policy module.

Please review.

zpytela:
Added typeattribute for backwards compatibility

wrabcak:
LGTM. 

Thanks.

PR #119, opened 2019-06-28
Add cyrus_delete_tmp() interface

zpytela:


ghost:
https://src.fedoraproject.org/rpms/cyrus-imapd/blob/master/f/cyrus-imapd.service#_12

This is PrivateTmp=true, Why not look into your options as to creating a more sustainablle solution.

you could for example consider creating a type attribute like systemd_private_tmp_type and then associate the (common) access systemd needs to PrivateTmp with the type attribute as well as associate PrivateTmp types with the systemd_private_tmp_type.

Example:
optional_policy(` systemd_private_tmp(cyrus_tmp_t) ')

Theres various benefits to this:

more efficient policy (less rules)
more readible policy (single point of failure, and its makes it pretty clear that this is for PrivateTmp=true)

wrabcak:
This is quite cool idea. It make sense. 

Let's focus on this solution.

ghost:
On Mon, Jul 01, 2019 at 11:51:53AM -0700, Lukas Vrabec wrote:
> This is quite cool idea. It make sense. 
> 
> Let's focus on this solution.
> 

So you'd get something like this:

systemd.te:

attribute systemd_private_tmp_type;

systemd.if:

######################################
## <summary>
##   Make the specified type usable as an
##   systemd private tmp type.
## </summary>
## <param name="domain">
##   <summary>
##     Type to be used as a private tmp type.
##   </summary>
## </param>
#
interface(`systemd_private_tmp',`
	gen_require(`
		attribute systemd_private_tmp_type;
	')

	typeattribute $1 systemd_private_tmp_type;
')

#######################################
## <summary>
##  Delete private tmp
## </summary>
## <param name="domain">
## <summary>
##  domain allowed access
## </summary>
## </param>
#
interface(`systemd_delete_private_tmp',`
	gen_require(`
		attribute systemd_private_tmp_type;
	')

	allow $1 systemd_private_tmp_type:dir delete_dir_perms)
	allow $1 systemd_private_tmp_type:fifo_file delete_fifo_file_perms)
	allow $1 systemd_private_tmp_type:file delete_file_perms)
	allow $1 systemd_private_tmp_type:lnk_file delete_lnk_file_perms)
	allow $1 systemd_private_tmp_type:sock_file delete_sock_file_perms)
')

init.te:

ifdef(`init_systemd',`
	# delete private tmp
	files_delete_tmp_dir_entry(init_t)
	systemd_delete_private_tmp(init_t)
')

cyrus.te:

optional_policy(`
	systemd_private_tmp(cyrus_tmp_t)
')

A lot of services have tmp content and have PrivateTmp=yes set in their systemd service unit.
You might want to fix up existing instances where this was allowed individually (go through init.te and see if init is allowed to delete tmp content, then see if that was allowed due to the existence of PrivateTmp=yes and fix accordingly)

You can apply this concept to other systemd directives as well. For example socket activation, PidFile=, PrivateDevices=, StateDirectory=, RuntimeDirectory= etc etc

It makes it much easier to sustain if you do it right. Also might also pay off if you ever intent to make an udica equivalent for systemd portables.

> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/fedora-selinux/selinux-policy-contrib/pull/119#issuecomment-507382297

-- 
Dominick Grift

PR #123, opened 2019-07-18
Allow domain transition lsmd_t to sensord_t

Koncpa:
Add macro sensord_domtrans(lsmd_t) which allow domain transition to sensord_t which
solve problem with bad transition where sensord-service get label lsmd_t because lsm daemon executes sensord
Resolves: rhbz#1662062

### DIFF
--- a/lsm.te
+++ b/lsm.te
@@ -55,6 +55,10 @@ optional_policy(`
     rpm_debuginfo_domtrans(lsmd_t)
 ')
 
+optional_policy(`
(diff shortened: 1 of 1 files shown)


Koncpa:
Fixes bug: https://bugzilla.redhat.com/show_bug.cgi?id=1662062

wrabcak:
Looks good to me, 

@Koncpa Thanks for contribution!

PR #128, opened 2019-07-25
Allow httpd_t to signull mailman_cgi_t process

5umm3r15:
Allow httpd_t to test for existence of mailman_cgi_t process without sending a signal.

Fixed Red Hat Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1686462#

### DIFF
--- a/apache.te
+++ b/apache.te
@@ -1024,6 +1024,7 @@ optional_policy(`
 
 optional_policy(`
 	mailman_signal_cgi(httpd_t)
+	mailman_signull_cgi(httpd_t)
 	mailman_domtrans_cgi(httpd_t)
 	mailman_read_data_files(httpd_t)
 	# should have separate types for public and private archives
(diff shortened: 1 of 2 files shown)


ghost:
this probably applies to all "httpd_script_type".  Probably eventually best off by allowing instead atlest:

`allow httpd_t httpd_script_type:process {signal signull sigkill};`

wrabcak:
Hi @doverride , 

Allow rule mentioned in your comment is already allowed in selinux-policy repository:
https://github.com/fedora-selinux/selinux-policy-contrib/blob/rawhide/apache.te#L1758

Maybe it make sense to redesign mailman_cgi_t domain to use apache_content_alias_template() template. We could focus on this in future, for know I'm merging this ticket. 

@5umm3r15 Thanks for contribution!

PR #130, opened 2019-08-12
cockpit: Allow cockpit-session to read cockpit-tls state

martinpitt:
Commit 819c407b87 was missing the permission for cockpit-session to read
cockpit-tls' validated certificates in /run/cockpit/tls/.

----

I reworked the old https://github.com/cockpit-project/cockpit/pull/11421 to now run against the new cockpit-tls. Without this, it fails with
```
AVC avc:  denied  { open } for  pid=8136 comm="cockpit-session" path="/run/cockpit/tls/ws.8129.crt" dev="tmpfs" ino=46659 scontext=system_u:system_r:cockpit_session_t:s0 tcontext=system_u:object_r:cockpit_var_run_t:s0 tclass=file permissive=0
AVC avc:  denied  { read } for  pid=11409 comm="cockpit-session" name="ws.11407.crt" dev="tmpfs" ino=72371 scontext=system_u:system_r:cockpit_session_t:s0 tcontext=system_u:object_r:cockpit_var_run_t:s0 tclass=file permissive=0
```

### DIFF
--- a/cockpit.te
+++ b/cockpit.te
@@ -142,6 +142,8 @@ manage_dirs_pattern(cockpit_session_t, cockpit_tmpfs_t, cockpit_tmpfs_t)
 manage_files_pattern(cockpit_session_t, cockpit_tmpfs_t, cockpit_tmpfs_t)
 fs_tmpfs_filetrans(cockpit_session_t, cockpit_tmpfs_t, { file })
 
+read_files_pattern(cockpit_session_t, cockpit_var_run_t, cockpit_var_run_t)
+
 kernel_read_network_state(cockpit_session_t)
 
 # cockpit-session runs a full pam stack, including pam_selinux.so


wrabcak:
LGTM. Merging. Should I backport it also to F30? 

Thanks,
Lukas.

martinpitt:
@wrabcak : Yes, please, that would be great. We'll also need it in RHEL 8.2, but I suppose I'll send a bugzilla about that. There will (most likely) be one more change coming when I let cockpit-tls and cockpit-ws run as different user IDs, so that they run in different security contexts, after that I'll ask for a RHEL update.

wrabcak:
@martinpitt , backported also to F30. For RHEL-8.2 please create bugzilla ticket, but there is a no problem to backport it there from technical POV.

Thanks for help!
Lukas.

PR #141, opened 2019-09-19
Run ipa-custodia as ipa_custodia_t

tiran:
Create policy files and rules for ipa-custodia service.
ipa-custodia is a helper service for FreeIPA to share secrets between
IPA servers.

Co-Authored-By: Lukas Vrabec <lvrabec@redhat.com>

### DIFF
--- a/apache.if
+++ b/apache.if
@@ -1832,6 +1832,27 @@ interface(`apache_read_pid_files',`
 
 ########################################
 ## <summary>
+##	Manage apache pid objects.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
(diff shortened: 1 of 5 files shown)


lslebodn:
Could you explain why it cannot be part of custodia package in fedora?
There is already upstream for that.
https://github.com/latchset/ipa-custodia-selinux/

And there are already packages in fedora custom selinux-policy
e.g. https://src.fedoraproject.org/rpms/ejabberd/blob/master/f/ejabberd.spec#_153

wrabcak:
@lslebodn it's too late for Fedora 31. 

Let's drop All IPA related stuff in Fedora 32.

wrabcak:
LGTM.

PR #148, opened 2019-10-05
nova.fc: fix duplicated slash

mikhailnov:


5umm3r15:
LGTM.

Thanks for contribution!

Nikola

PR #151, opened 2019-10-11
Allow ipmievd_t to RW watchdog devices

5umm3r15:
Allow ipmievd - IPMI event daemon for sending events to syslog to read/write to watchdog devices.


wrabcak:
LGTM. thanks.

PR #152, opened 2019-10-11
Update numad policy to allow signull, kill, nice and trace, processes

Koncpa:
Numad is user-level daemon that provides placement advice and process
management for efficient use of CPUs and memory on systems with NUMA topology
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1759561

Allow capability sys_nice set real-time scheduling policies for calling process,
and set scheduling policies and priorities for arbitrary processes
Allow capability sys_ptrace to trace arbitrary processes
Allow capability kill to kill processes
Allow macro domain_signull_all_domains() to send a null signal to all domains

### DIFF
--- a/numad.te
+++ b/numad.te
@@ -23,7 +23,7 @@ files_pid_file(numad_var_run_t)
 # numad local policy
 #
 
-allow numad_t self:capability sys_ptrace;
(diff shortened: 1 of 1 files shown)


wrabcak:
LGTM. 
Thanks.

PR #156, opened 2019-10-22
Update cockpit policy

Koncpa:
Allow to systemd to use dir with file context cockpit_var_run_t as mount point
Allow SELinux Domain trasition from sytemd into confined domain with NoNewPrivileges
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1761765


wrabcak:
LGTM.

PR #158, opened 2019-10-25
Allow capability dac_override in logwatch_mail_t domain 

Koncpa:
Allow capability dac_override in logwatch_mail_t domain, which allow bypass file read, write, and execute permission checks.
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1761072

### DIFF
--- a/logwatch.te
+++ b/logwatch.te
@@ -188,7 +188,7 @@ optional_policy(`
 # Mail local policy
 #
 
-allow logwatch_mail_t self:capability { dac_read_search  };
+allow logwatch_mail_t self:capability { dac_override };
 
 allow logwatch_mail_t logwatch_t:fd use;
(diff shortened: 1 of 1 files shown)


WOnder93:
> Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1761072

This BZ is not accessible from the public. Please replace the link with an explanation why logwatch_mail_t needs this capability (e.g. what program/service under what configuration/scenario breaks without it).

Koncpa:
Now is bug not private, so you can see more detail why logwatch_mail_t need specific capabilty.

wrabcak:
Bugzilla ticket is now public so let's keep reference bug in commit msg. Logwatch maintainers doesn't care about SELinux policy, so let's just allow it.

PR #161, opened 2019-11-04
cockpit: Support https instance factory

martinpitt:
https://github.com/cockpit-project/cockpit/pull/12972 hardens Cockpit's
web server so that each different client certificate gets handled by its
own cockpit-ws instance. This requires a factory helper that produces
new systemd template instances.

This completes the TLS rework from commit 819c407b8744d25.


martinpitt:
It would be great if this could land in F30 and F31 (and eventually in RHEL 8.2, but I'll file a bugzilla for that). Shall I send separate PRs for these branches, or do you merge stuff in bulk?

Thank you!

wrabcak:
LGTM. 
I'll cherry-pick it to the F30 and F31.

PR #162, opened 2019-11-09
cockpit: Allow cockpit-session to read cockpit-tls state directory

martinpitt:
This is required for the current approach of cockpit-tls exporting one
certificate for every connection, which avoids error prone refcounts.
For that, the PAM module needs to glob for /run/cockpit/tls/<fingerprint>-*.

Resolves: rhbz#1770221


martinpitt:
As usual, it would be great if you can cherry-pick this into F30 and F31. Thanks!

wrabcak:
LGTM. Merging.

wrabcak:
cherry-picked to F31 and F30

PR #163, opened 2019-11-12
cockpit: Drop cockpit-cert-session

martinpitt:
This separate binary only existed in a proof of concept, but got dropped
later on (functionality merged into cockpit-session).

This reverts part of commit e5631ccdd200c.


martinpitt:
Note, this is just cleanup -- it doesn't break anything, so there is no hurry to backport this anywhere.

wrabcak:
@martinpitt , so do you agree with merging this for Rawhide only? (Future f32) 

Thanks,
Lukas.

martinpitt:
@wrabcak : Yes, that's fine. There's no such binary, so there's no harm in schlepping that extra rule for a while in F30/31.

wrabcak:
Okay, thank you for your contribution.

PR #175, opened 2019-12-06
Change type in transition for /var/cache/{dnf,yum} directory

zpytela:
In the rpm_named_filetrans interface calling files_var_filetrans()
for dnf and yum, change the target type from rpm_var_lib_t
to rpm_var_cache_t to match default file context.


zpytela:
Before:
```
f32# rpm -q selinux-policy; rmdir /var/cache/yum && mkdir /var/cache/yum && ls -dZ /var/cache/yum && matchpathcon /var/cache/yum
selinux-policy-3.14.5-18.fc32.noarch
unconfined_u:object_r:rpm_var_lib_t:s0 /var/cache/yum
/var/cache/yum  system_u:object_r:rpm_var_cache_t:s0
```
After:
```
f32# rpm -q selinux-policy; rmdir /var/cache/yum && mkdir /var/cache/yum && ls -dZ /var/cache/yum && matchpathcon /var/cache/yum
selinux-policy-3.14.5-18.200.fc32.noarch
unconfined_u:object_r:rpm_var_cache_t:s0 /var/cache/yum
/var/cache/yum  system_u:object_r:rpm_var_cache_t:s0
```
Please review the PR.

rpm.if
```diff
@@ -431,9 +431,8 @@ interface(`rpm_named_filetrans',`
 	logging_log_named_filetrans($1, rpm_log_t, file, "yum.log")
 	logging_log_named_filetrans($1, rpm_log_t, file, "hawkey.log")
 	logging_log_named_filetrans($1, rpm_log_t, file, "up2date")
-	files_var_filetrans($1, rpm_var_lib_t, dir, "dnf")
-	files_var_filetrans($1, rpm_var_lib_t, dir, "yum")
-	files_var_filetrans($1, rpm_var_lib_t, dir, "rpm")
```
wrabcak, line 436: Why file named transition for "rpm" directory is missing?

wrabcak (changes requested):

rpm.if
zpytela: Omitted intentionally as there rpm does not seem to use any other directory in /var than /var/lib/rpm which is confined, namely there is no /var/cache/rpm. The var_t type is assigned to /var and /var/.* when no other particular type is assigned.

wrabcak: Understand, but it would be great to discuss it with RPM people to be sure. We want to avoid regressions.

zpytela: I raised the question. I doubt though as we don't have the support in rpm.fc.

wrabcak (approved):

wrabcak:
Let's merge it and based on their response we could improve the fix in future. 
THanks.

zpytela:
@wrabcak FYI confirmed: no /var/cache/rpm since the turn of millenium :)

PR #185, opened 2020-01-15
Allow opafm_t to create and use netlink rdma sockets.

5umm3r15:
Allow opafm_t-Intel Omni-Path Fabric Host Software, to create and use netlink_rdma (remote direct memory access) socket.

Intel Omni-Path Fabric Host Software installs components needed to set up compute, I/O, and Service nodes with drivers, stacks, and basic tools for local configuration and monitoring.

Fixed Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1786670


wrabcak:
LGTM.

PR #188, opened 2020-01-27
Allow thumb_t connect and send to system_dbusd_t BZ(1795044)

zpytela:


thumb.te
```diff
@@ -143,6 +143,8 @@ optional_policy(`
     dbus_connect_session_bus(thumb_t)
 	dbus_stream_connect_session_bus(thumb_t)
 	dbus_chat_session_bus(thumb_t)
+	dbus_stream_connect_system_dbusd(thumb_t)
```
wrabcak: Why is this rule needed when you also adding macro "dbus_system_bus_client(thumb_t)" ?

wrabcak (changes requested):

thumb.te
zpytela: I was running off other policies containing both the macros so I haven't realize the second one already contains stream_connect_pattern().
Changed.

wrabcak (approved):

PR #196, opened 2020-02-03
Label /run/user/.../bus as session_dbusd_tmp_t

secureworkstation:
I need that in order to make clearer permissions in my KDE patchset.


zpytela:
@secureworkstation, can this line be changed to match a socket file only?
/var/run/user/[^/]*/bus

PR #197, opened 2020-02-03
Label /run/user/.../pulse/ as pulseaudio_home_t

secureworkstation:


pulseaudio.if
```diff
@@ -293,6 +297,43 @@ interface(`pulseaudio_manage_home_files',`
...
+##      </summary>
+## </param>
+#
+interface(`pulseaudio_manage_session_tmp_files',`
+        gen_require(`
+                type pulseaudio_home_t;
+        ')
+
+        manage_files_pattern($1, pulseaudio_home_t, pulseaudio_home_t)
+        allow $1 pulseaudio_home_t:sock_file { create unlink };
```
zpytela, line 334: Are these permissions sufficient? Domains like systemd_logind_t have some generic permissions which may not be true for other domains.

zpytela:
@secureworkstation, i suggest changing '*' to '+' to match non-empty string only:
/var/run/user/([^/]*)/pulse(/.*)?

PR #198, opened 2020-02-04
Allow networkmanager_t execute setfiles_t

zpytela:
Resolves: rhbz#1787162


wrabcak:
@zpytela 
Could we use "seutil_run_setfiles" rather then directly allow NetworkManager_t to relabel non security files?

zpytela:
Changed.

PR #201, opened 2020-02-06
Add qemu_read_pid_files() interface

zpytela:


virt.if
```diff
@@ -1759,3 +1777,22 @@ interface(`virt_svirt_manage_tmp',`
...
+## <summary>
+##     Read qemu PID files.
+## </summary>
+## <param name="domain">
+##     <summary>
+##     Domain allowed access.
+##     </summary>
+## </param>
+#
+interface(`qemu_read_pid_files',`
```
wrabcak: Let's use "virt" prefix for interface name to make clear in which interface file is the interface declared.

virt.if
```diff
@@ -1539,6 +1539,24 @@ interface(`virt_rw_svirt_image',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`virt_write_svirt_image_sock_file',`
+	gen_require(`
+		type svirt_image_t;
+	')
+
+	write_sock_files_pattern($1, svirt_image_t, svirt_image_t)
```
wrabcak, line 1557: Do we know which objects are labeled as svirt_image_t ? I'm not saying it's wrong, just want to understand if the svirt_image_t for sockets is good approach.

wrabcak (changes requested):

wrabcak (approved):

PR #202, opened 2020-02-06
Allow vhostmd communication with hosted virtual machines

zpytela:
Allow vhostmd_t domain qemu_read_pid_files(), virt_stream_connect_svirt(),
and virt_write_svirt_image_sock_file().


zpytela:
Requires interfaces introduced in:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/201

PR #204, opened 2020-02-10
Allow bacula dac_override capability

zpytela:
The dac_override capability is needed to restore permissions and ownership.
Resolves: rhbz#1741609


zpytela:
This should be backported to F31 and F30, too.

wrabcak:
LGTM.

PR #206, opened 2020-02-13
Update policy for NetworkManager_ssh_t

5umm3r15:
corenet_rw_tun_tap_dev: Allow NetworkManager_ssh_t to read write tun_tap_device chr_files

init_rw_stream_sockets:  Allow NetworkManager_ssh_t to read/write to init with a unix domain stream sockets.

kernel_search_network_sysctl: Allow NetworkManager_ssh_t to search network sysctl directories

kernel_request_load_module: Allow NetworkManager_ssh_t to request to load kernel module

sysnet_domtrans_ifconfig: Allow NetworkManager_ssh_t to execute ifconfig in the ifconfig domain

t̶e̶r̶m̶_̶u̶s̶e̶_̶p̶t̶m̶x̶:̶ ̶A̶l̶l̶o̶w̶ ̶N̶e̶t̶w̶o̶r̶k̶M̶a̶n̶a̶g̶e̶r̶_̶s̶s̶h̶_̶t̶ ̶t̶o̶ ̶r̶e̶a̶d̶ ̶a̶n̶d̶ ̶w̶r̶i̶t̶e̶ ̶t̶h̶e̶ ̶p̶t̶y̶ ̶m̶u̶l̶t̶i̶p̶l̶e̶x̶o̶r̶ ̶(̶/̶d̶e̶v̶/̶p̶t̶m̶x̶)̶.̶

Fixed Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1761071

### DIFF
--- a/networkmanager.te
+++ b/networkmanager.te
@@ -449,12 +449,13 @@ optional_policy(`
 
 optional_policy(`
     ssh_basic_client_template(NetworkManager, NetworkManager_t, system_r)
-    ssh_exec(NetworkManager_ssh_t)
-    term_use_generic_ptys(NetworkManager_ssh_t)
     modutils_domtrans_kmod(NetworkManager_ssh_t)
+    corenet_rw_tun_tap_dev(NetworkManager_ssh_t)
     dbus_connect_system_bus(NetworkManager_ssh_t)
     dbus_system_bus_client(NetworkManager_ssh_t)
+    init_rw_stream_sockets(NetworkManager_ssh_t)
     networkmanager_dbus_chat(NetworkManager_ssh_t)
+    sysnet_domtrans_ifconfig(NetworkManager_ssh_t)
 ')
 
 optional_policy(`


ghost:
I haven't looked at the enclosed logs with the bugzilla but just to note:

`term_use_ptmx` is generally not meant to be called directly. This event happens when the caller creates a pty, and ptys should generally not end up with the same label as /dev/pts (devpts_t i guess, as this is a filesystem type). Therefore depending on what
the created pty is used for either one should declare a new type for "NetworkManager SSH pty's" and then use `term_create_pty()` to allow NetworkManager_ssh_t to create ptys with the specified pty type. Or one should allow NetworkManager_ssh_t to create userdom user pty's is this pty is to be used by users (i guess this is unlikely). The `term_create_pty()` interface takes care of the `term_use_ptmx()` requirement.

5umm3r15:
Hi @doverride

Thanks for explaining.
SELinux denials where NetworkManager_ssh_t was prevented from managing ptmx_t files happens after reproducing this bug using strace and tmux.
So this macro was not necessary to be applied and I have removed it. 

Nikola

networkmanager.te
```diff
@@ -443,10 +443,14 @@ optional_policy(`
     ssh_basic_client_template(NetworkManager, NetworkManager_t, system_r)
     term_use_generic_ptys(NetworkManager_ssh_t)
     modutils_domtrans_kmod(NetworkManager_ssh_t)
+    corenet_rw_tun_tap_dev(NetworkManager_ssh_t)
     dbus_connect_system_bus(NetworkManager_ssh_t)
     dbus_system_bus_client(NetworkManager_ssh_t)
-
+    init_rw_stream_sockets(NetworkManager_ssh_t)
+    kernel_request_load_module(NetworkManager_ssh_t)
```
zpytela: when ifconfig transitions to its domain, is this permission still required?

networkmanager.te
```diff
@@ -443,10 +443,14 @@ optional_policy(`
     ssh_basic_client_template(NetworkManager, NetworkManager_t, system_r)
     term_use_generic_ptys(NetworkManager_ssh_t)
     modutils_domtrans_kmod(NetworkManager_ssh_t)
+    corenet_rw_tun_tap_dev(NetworkManager_ssh_t)
     dbus_connect_system_bus(NetworkManager_ssh_t)
     dbus_system_bus_client(NetworkManager_ssh_t)
-
+    init_rw_stream_sockets(NetworkManager_ssh_t)
```
zpytela, line 456: In the avc list, I see only getattr and ioctl reported. Did you reproduce it so that also other permissions were requested?

zpytela (commented):
@5umm3r15, 2 comments inline.

networkmanager.te
ghost: this is a common access:

NetworkManager_ssh_t init_t:fd use;
NetworkManager_ssh_t init_t:unix_stream_socket { getattr ioctl read write };

all processes forked by systemd (directly or indirectly) need this. In this case networkmanager_ssh_t is probably executed by networkmanager_t which is executed by init_t, so even though networkmanager_ssh_t is not a direct decendant of init_t it still is indirect. I believe this event is triggered by the way systemd/journald intercept stdout (logging "feature")

networkmanager.te
5umm3r15: Thanks @zpytela, you are right. 
I removed interfaces: kernel_search_network_sysctl and kernel_request_load_module, then tried it with scratch build and this permissions are not required.

networkmanager.te
5umm3r15: Thanks @doverride

zpytela:
@5umm3r15 I suppose you need to rebase the commit now as networkmanager.te has changed in the meantime.

5umm3r15:
@zpytela conflicts are resolved, can you please review it? 
Thanks

zpytela:
@5umm3r15, the content LGTM, I don't quite get though why there are 2 commits.

5umm3r15:
Hi @zpytela , I removed one commit and rebased it.

zpytela:
LGTM.

PR #207, opened 2020-02-14
Update radiusd policy

terryburton:
Allow SELinux Domain trasition from systemd into confined domain with NoNewPrivileges

Allow systemd to use dir with file contexts radiusd_etc_t, radiusd_log_t and
radiusd_var_run_t as mount points

Upstream FreeRADIUS systemd unit has NNP and private mount spaces:

NoNewPrivileges=true
RuntimeDirectory=radiusd
PrivateTmp=true
ReadOnlyDirectories=/etc/raddb/
ReadWriteDirectories=/var/log/radius/

Fedora will likely adopt this.

### DIFF
--- a/radius.te
+++ b/radius.te
@@ -15,9 +15,11 @@ gen_tunable(radius_use_jit, false)
 type radiusd_t;
 type radiusd_exec_t;
 init_daemon_domain(radiusd_t, radiusd_exec_t)
+init_nnp_daemon_domain(radiusd_t)
 
 type radiusd_etc_t;
 files_config_file(radiusd_etc_t)
+systemd_mount_dir(radiusd_etc_t)
 
 type radiusd_etc_rw_t;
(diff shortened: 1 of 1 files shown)


wrabcak:
Hi @terryburton , 

Thanks for your PR and making radius policy synced with upstream. :) 
Should I backport it to stable versions of Fedora or Rawhide is enough for now? 

Thanks,
Lukas.

terryburton:
@wrabcak Thanks for the quick merge.

The upstream packages do not currently ship with any SELinux policy, but do have a restrictive systemd unit that triggers AVC denies with the default SELinux policies shipped with distros. Users report issues on recent CentOS and RHEL, for example.

Therefore backporting this to supported Fedora version would be appreciated. Thank you!

wrabcak:
Patch backported to F31 and F30. 

Thanks,
Lukas.

PR #179, opened 2019-12-17
Add new file context rabbitmq_conf_t.

Richard-Filo:
The rabbitmq_conf_t is context for configuration files.
Allow rabbitmq_t domain to manage files and dirs labled rabbitmq_conf_t.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1780824

PR #210, opened 2020-02-24
Revert "Add new file context rabbitmq_conf_t."

Richard-Filo:
Reverts fedora-selinux/selinux-policy-contrib#179

Reason is that I forgot to ask maintainer of rabbitmq if it is expected behavior that a program which rabbitmq server uses, can write to own configuration file.
I create BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1803126.
And now I'm waiting for answer. In the meantime I'm reverting this PR.

### DIFF
--- a/rabbitmq.fc
+++ b/rabbitmq.fc
@@ -9,5 +9,3 @@
 /var/log/rabbitmq(/.*)?	gen_context(system_u:object_r:rabbitmq_var_log_t,s0)
 
 /var/run/rabbitmq(/.*)?	gen_context(system_u:object_r:rabbitmq_var_run_t,s0)
-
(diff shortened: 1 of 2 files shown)


zpytela:
np with reverting; but what will be the next step?

wrabcak:
It's important also to understand, why we're reverting it and avoid it in future.

WOnder93:
Also please explain the reason for reverting in the commit message. Revert without any explanation is a very bad practice.

zpytela:
@Richard-Filo, please  include the explanation to the commit message so that it is clear and understandable and make the first comment match the commit message.

PR #211, opened 2020-02-26
Allow postfix stream connect to cyrus through runtime socket

zpytela:
Add cyrus_runtime_stream_connect() interface.
Allow postfix_smtp_t connect to cyrus_t through a socket in the runtime
filesystem using the cyrus_runtime_stream_connect() interface.

Resolves: rhbz#1807443

### DIFF
--- a/cyrus.if
+++ b/cyrus.if
@@ -61,6 +61,26 @@ interface(`cyrus_stream_connect',`
 
 ########################################
 ## <summary>
+##	Connect to Cyrus using a unix
+##	domain stream socket in the runtime filesystem.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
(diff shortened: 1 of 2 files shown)


wrabcak:
Does it make sense to merge cyrus_stream_connect and cyrus_runtime_stream_connect interfaces?

zpytela:
I was considering it, but currently it seems to be required for postfix_smtp_t only.

wrabcak:
Do we know what are differences between these two sockets?

zpytela:
The lmtp one is for local mail delivery. In /var/lib/imap, there are multiple ones for e. g. imapd or pop3d and I am not fully aware of their purpose.

wrabcak:
Ok Let's merge it as it is.

PR #215, opened 2020-03-09
Allow sssd read systemd-resolved runtime directory

zpytela:
Resolves: rhbz#1811556

### DIFF
--- a/sssd.te
+++ b/sssd.te
@@ -192,6 +192,7 @@ optional_policy(`
 
 optional_policy(`
 	systemd_login_read_pid_files(sssd_t)
+	systemd_resolved_read_pid(sssd_t)
 ')
 
 optional_policy(`


wrabcak:
Is it connected to this issue? https://github.com/fedora-selinux/selinux-policy-contrib/commit/6ac41c5b089731d9c5509db5ef6eae5944263b20 

If yes, would it make sense to make one macro putting these allow rules together?

zpytela:
If accepted, it should be backported also to F31 and F30.

@wrabcak, I was rather questioning myself if both these permission should be allowed for nsswitch_domain. So far it was only requested for sssd_t though.

wrabcak:
Maybe we should be proactive here and address it before users start reporting bugzilla tickets. 

@lslebodn what is your opinion here?

lslebodn:
I assume rhbz#1811556 is related to reading file /run/systemd/resolve/resolv.conf

```
sh$ semanage fcontext -l | grep systemd_resolved_var_run_t
/run/systemd/resolve(/.*)?                         all files          system_u:object_r:systemd_resolved_var_run_t:s0 
/var/run/systemd/resolve(/.*)?                     all files          system_u:object_r:systemd_resolved_var_run_t:s0
```

and thus it should be allowed just for sssd daemon itself `sssd_t`. It is not related to the client code.

zpytela:
@wrabcak, given the answer I propose the PR for review as is.

PR #216, opened 2020-03-10
Create geoclue_t domain with init_daemon_domain()

zpytela:
Create geoclue_t domain with init_daemon_domain() interface instead of
application_domain() as it matches the "long running processes" condition
rather than "programs that are run interactively".

Resolves: rhbz# 1795524


wrabcak:
I'm not sure if it's good idea to remove application_domain, we can remove some rules by this change and regress. I would add just the macro init_daemon_domain()

zpytela:
Changed. There is a small set of different rules only, but they are there.

PR #218, opened 2020-03-12
Allow NetworkManager_t getattr NetworkManager_unit_file_t

zpytela:


wrabcak:
I'm afraid this won't be enough of allow rules. Most probably we need at least read permissions for NetworkManager_unit_file_t  for networkmanager_t and maybe rule like this:
allow NetworkManager_t NetworkManager_unit_file_t:service manage_service_perms;

zpytela:
In permissive mode, only getattr appears. The only command executed in the NetworkManager_t domain seems to be
```
systemctl -q is-enabled nm-cloud-setup.service
```

wrabcak:
Can somebody confirm it from network manager community ?

zpytela:
Looking at:
```
  $ cat /usr/lib/NetworkManager/dispatcher.d/no-wait.d/90-nm-cloud-setup.sh 
#!/bin/sh

case "$2" in
    up|dhcp4-change)
        if systemctl -q is-enabled nm-cloud-setup.service ; then
            exec systemctl --no-block restart nm-cloud-setup.service
        fi
        ;;
esac
```
it seems the test does not cover this part, so adding the required rules. Anyway, what we actually need is functional testing from the cloud-setup team which has already been requested.

wrabcak:
Perfect. Thanks. :)

PR #222, opened 2020-03-19
Allow targetd to exec lvm command

tasleson:
Later versions of lvm removed the lvm2app API which targetd was
using.  Targetd has switched over to use libblockdev for its lvm
needs.  Thus targetd needs the ability to fork & exec lvm binary
which is what libblockdev does when using the lvm plugin.

Additionally the `lvm` command needs access to `/run/lvm/hints` and the ability to send dbus messages.

### DIFF
--- a/targetd.te
+++ b/targetd.te
@@ -5,6 +5,13 @@ policy_module(targetd, 1.0.0)
 # Declarations
 #
 
+require {
+    type lvm_t;
+    type lvm_var_run_t;
+    class file { getattr read write open lock };
+    class dbus send_msg;
+}
+
 type targetd_t;
 type targetd_exec_t;
 init_daemon_domain(targetd_t, targetd_exec_t)
@@ -23,6 +30,8 @@ files_tmp_file(targetd_tmp_t)
 # targetd local policy
 #
 
+
+
 allow targetd_t self:capability { ipc_lock sys_admin sys_nice };
 allow targetd_t self:fifo_file rw_fifo_file_perms;
 allow targetd_t self:unix_stream_socket create_stream_socket_perms;
@@ -30,6 +39,8 @@ allow targetd_t self:unix_dgram_socket create_socket_perms;
 allow targetd_t self:tcp_socket { accept listen };
 allow targetd_t self:netlink_route_socket r_netlink_socket_perms;
 allow targetd_t self:process { setfscreate setsched };
+allow targetd_t lvm_t:dbus send_msg;
+allow targetd_t lvm_var_run_t:file { getattr read write open lock };
 
 manage_dirs_pattern(targetd_t, targetd_etc_rw_t, targetd_etc_rw_t)
 manage_files_pattern(targetd_t, targetd_etc_rw_t, targetd_etc_rw_t)
@@ -93,6 +104,7 @@ optional_policy(`
    lvm_manage_lock(targetd_t)
    lvm_rw_pipes(targetd_t)
    lvm_stream_connect(targetd_t)
+   lvm_exec(targetd_t)
 ')
 
 optional_policy(`


zpytela:
Tony,

Thank you for this PR. There need to be, however, 2 new interfaces created in lvm.if first to access types from a different module, in a similar way how lvm_exec is used in the commit.

Instead of enumerating permissions, macros like `write_file_perms` can be used.

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/lvm.if
https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt

tasleson:
> Tony,
> 
> Thank you for this PR. There need to be, however, 2 new interfaces created in lvm.if first to access types from a different module, in a similar way how lvm_exec is used in the commit.

OK, please elaborate here a bit.  Are you saying I need to add a couple new interface to `lvm.if` so that the targetd process can utilize the lvm dbus send and access to the hints file?

I was actually surprised that I needed to add permissions for the lvm binary to do a dbus send and access lvm hints file as those should already be included in `lvm_exec` IMHO.

> 
> Instead of enumerating permissions, macros like `write_file_perms` can be used.

I don't see a single macro that has the exact permissions needed.  So do I pick one that has more than is needed to give more attack surface area or create a new macro?

> 
> https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/lvm.if
> https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt

zpytela:
> > Tony,
> > Thank you for this PR. There need to be, however, 2 new interfaces created in lvm.if first to access types from a different module, in a similar way how lvm_exec is used in the commit.
> 
> OK, please elaborate here a bit. Are you saying I need to add a couple new interface to `lvm.if` so that the targetd process can utilize the lvm dbus send and access to the hints file?

Exactly. The lvm module contains the definition of types, attributes, interfaces. The interfaces can be used outside of this module for usage by other modules.

> 
> I was actually surprised that I needed to add permissions for the lvm binary to do a dbus send and access lvm hints file as those should already be included in `lvm_exec` IMHO.

See the interface content in lvm.if or use macro-expander command to see what an interface allows. In the interface heading, you can see:
```
Execute lvm programs in the caller domain
```

> > Instead of enumerating permissions, macros like `write_file_perms` can be used.
> 
> I don't see a single macro that has the exact permissions needed. So do I pick one that has more than is needed to give more attack surface area or create a new macro?

A rule of thumb is that if you are sure enough no more permissions will be needed in the near future, use your own set, otherwise choose the closest superset.

> 
> > https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/lvm.if
> > https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/support/obj_perm_sets.spt

tasleson:
> See the interface content in lvm.if or use macro-expander command to see what an interface allows. In the interface heading, you can see:

While looking at `lvm.if`, to do some editing I found `lvm_run` which states:

```
Execute lvm programs in the lvm domain.
```

Which I think I should be using instead of `lvm_exec`.  `lvm_run` is being used by `mount.te`.  However, when I simply edit `targetd.te` to use `lvm_run` instead, I get the following error.

```diff
diff --git a/targetd.te b/targetd.te
index b7f5164d3..72dc3bfb4 100644
--- a/targetd.te
+++ b/targetd.te
@@ -93,6 +93,7 @@ optional_policy(`
    lvm_manage_lock(targetd_t)
    lvm_rw_pipes(targetd_t)
    lvm_stream_connect(targetd_t)
+   lvm_run(targetd_t)
 ')
```
```
make -f /usr/share/selinux/devel/Makefile targetd.pp
Compiling targeted targetd module
targetd.te:88:ERROR 'syntax error' at token 'types' on line 6652:
	role  types lvm_t;
#line 88
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
```

Which at the moment, I'm not sure why?

@zpytela Suggestions?

tasleson:
@zpytela I'm trying to follow your guidelines.  I can get the following change to compile and load (It's not done), but how do I get changes to targetd to use the new interfaces to compile and load it?  What is needed to provide the new interfaces for others to consume?

```patch
diff --git a/policy/modules/system/lvm.if b/policy/modules/system/lvm.if
index fbbb39e5a..9f7ee32dc 100644
--- a/policy/modules/system/lvm.if
+++ b/policy/modules/system/lvm.if
@@ -452,4 +452,41 @@ interface(`lvm_manage_lock',`
 ')
 
 
+########################################
+## <summary>
+##     Allow dbus send for lvm dbus API
+## </summary>
+## <param name="domain">
+##     <summary>
+##     Domain allowed access.
+##     </summary>
+## </param>
+#
+interface(`lvm_dbus_send_msg',`
+       gen_require(`
+               type lvm_t;
+               class dbus send_msg;
+       ')
+    allow $1 lvm_t:dbus send_msg
+')
+
+########################################
+## <summary>
+##     Allow lvm hints file access
+## </summary>
+## <param name="domain">
+##     <summary>
+##     Domain allowed access.
+##     </summary>
+## </param>
+#
+interface(`lvm_hints_file_access',`
+       gen_require(`
+               type lvm_t;
+               type lvm_var_run_t;
+               class file { getattr read write open lock };
+       ')
+       allow $1 lvm_var_run_t:file { getattr read write open lock };
+')
+

```

I haven't found any developer docs which explain this, but I might not be looking in the right locations.

tasleson:
superseded with this PR: https://github.com/fedora-selinux/selinux-policy-contrib/pull/334

PR #223, opened 2020-03-23
Merge ipa and ipa_custodia modules

vmojzis:
ipa-custodia is an internal service for IPA. The policy for ipa-custodia was
separate to simplify it's transition under IPA maintenance, but it is no longer
needed since IPA adopted the whole policy module.
Since the two modules got merged in downstream IPA, maintaining separate
ipa_custodia module here only complicates installing the updated module by
freeipa-selinux package (it needs to disable ipa_custodia before installing the
updated module).


vmojzis:
Rebased.

ipa.fc
```diff
@@ -27,3 +27,10 @@
 
 /var/run/ipa(/.*)?              gen_context(system_u:object_r:ipa_var_run_t,s0)
 
+/usr/libexec/ipa/ipa-custodia					--	gen_context(system_u:object_r:ipa_custodia_exec_t,s0)
+/usr/libexec/ipa/custodia/ipa-custodia-dmldap			--	gen_context(system_u:object_r:ipa_custodia_dmldap_exec_t,s0)
+/usr/libexec/ipa/custodia/ipa-custodia-pki-tomcat		--	gen_context(system_u:object_r:ipa_custodia_pki_tomcat_exec_t,s0)
+/usr/libexec/ipa/custodia/ipa-custodia-pki-tomcat-wrapped	--	gen_context(system_u:object_r:ipa_custodia_pki_tomcat_exec_t,s0)
+/usr/libexec/ipa/custodia/ipa-custodia-ra-agent		--	gen_context(system_u:object_r:ipa_custodia_ra_agent_exec_t,s0)
+
+/var/log/ipa-custodia.audit.log(/.*)?				--	gen_context(system_u:object_r:ipa_custodia_log_t,s0)
```
zpytela, line 36: Is this intentional? Path regex matching both file and directory, but "--" for file spec. I see it was previously there, too.

ipa.if
```diff
@@ -308,3 +308,62 @@ interface(`ipa_read_tmp',`
 
     read_files_pattern($1, ipa_tmp_t, ipa_tmp_t)
 ')
+
+## <summary>policy for ipa_custodia</summary>
```
zpytela: I suppose the original header can be omitted here.

ipa.if
```diff
@@ -308,3 +308,62 @@ interface(`ipa_read_tmp',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`ipa_custodia_stream_connect',`
+	gen_require(`
+		type ipa_custodia_t;
+	')
+
+    allow $1 ipa_custodia_t:unix_stream_socket { connectto };
```
zpytela: I would take this as an opportunity to adjust formatting of this line to match the convention in the rest of the file.

zpytela:
If I read it right, it was copied from ipa_custodia files with some formatting adjustments (see a few comments inline), and also with adding 4 blocks of interfaces calls. @vmojzis, were these blocks added after checking with ipa engineers?

ipa.fc
vmojzis: Don't know about this one, I just moved it from ipa_custodia.fc, but it makes sence to fix it. Just not sure if it's better to exclude the directroy (to keep current labeling), or remove "--".

ipa.if
vmojzis: Sure, I thought it would denote "ipa_custodia" section, but apparently that is only done in ".te" files. I'll remove it.

ipa.if
vmojzis: The rest of the file doesn't really have a convention. It combines spaces and tabs quite a lot, which is probably why this indent is done in spaces instead of tabs (which I didn't notice since my editor set the indent lenght to 4 spaces). But I'll make it uniform. Thanks.

vmojzis:
> If I read it right, it was copied from ipa_custodia files with some formatting adjustments (see a few comments inline), and also with adding 4 blocks of interfaces calls. @vmojzis, were these blocks added after checking with ipa engineers?

Yes, I simply copied it over form ipa_custodia files and than fixed some formating, but I didn't add any interface calls.

vmojzis:
Updated and rebased. Thanks for the review @zpytela

zpytela:
> > If I read it right, it was copied from ipa_custodia files with some formatting adjustments (see a few comments inline), and also with adding 4 blocks of interfaces calls. @vmojzis, were these blocks added after checking with ipa engineers?
> 
> Yes, I simply copied it over form ipa_custodia files and than fixed some formating, but I didn't add any interface calls.
I see, sorry for misreading, the formatting change made me to get it wrong.

ipa.if
zpytela: The files do not follow a convention, but if at least adjacent blocks do, it makes it more readable.

ipa.fc
zpytela: @vmojzis, IMO makes sense to have specific type for the directory, too.

vmojzis: Fixed, but it changes the context of /var/log/ipa-custodia.audit.log directory (assuming ipa creates one) from var_log_t to ipa_custodia_log_t.

vmojzis:
Udpated.

zpytela:
@vmojzis LGTM now, thank you.

ipa.if
vmojzis: I'll cover this fix in a separate PR to ensure proper testing.

ipa.fc
vmojzis: I removed the fix again and will create a new PR for it to ensure proper testing.

PR #224, opened 2020-03-24
Add ibacm_t ipc_lock capability

zpytela:
Add ibacm_t the ipc_lock capability to support the update in kernel
to handle memory locking when interfacing with the Infiniband driver.

### DIFF
--- a/ibacm.te
+++ b/ibacm.te
@@ -25,6 +25,7 @@ files_tmpfs_file(ibacm_tmpfs_t)
 #
 # ibacm local policy
 #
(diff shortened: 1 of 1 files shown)


wrabcak:
LGTM.

wrabcak:
@zpytela 
Should I back ported to Fedora 31?

zpytela:
@wrabcak, the fix is in kernel since 201809, so anybody using ib will hit the issue.

wrabcak:
It looks like answer to my question is yes, so backporting to Fedora 31 and 30.

PR #227, opened 2020-03-27
Update ninfod policy to add nnp transition from systemd to ninfod

Koncpa:
Allow SELinux Domain trasition from systemd into confined domain with NoNewPrivileges
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1808987
Fedora COPR: https://copr.fedorainfracloud.org/coprs/pkoncity/selinux-policy/build/1313518/


wrabcak:
LGTM.

PR #231, opened 2020-04-01
Allow NetworkManager_ssh_t to execute_no_trans for binary ssh_exec_t

zpytela:
Add can_exec macro for NetworkManager_ssh_t domain, which allow to execute_no_trans for binary with label ssh_exec_t.
Bugzilla:  https://bugzilla.redhat.com/show_bug.cgi?id=1813098


wrabcak:
Merging because this is just backport from rawhide.

@zpytela  OT: we need to back-port also fixes from container-selinux because F31 is also affected by that bug.

PR #233, opened 2020-04-06
Allow ssh-keygen create file in /var/lib/glusterd

zpytela:
Gluster creates new ssh key pair in /var/lib/glusterd labeled as
glusterd_var_lib_t. Therefore ssh-keygen labeled as ssh_keygen_t needs to
have access to that directory.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1813917


wrabcak:
LGTM

PR #234, opened 2020-04-09
Allow rngd create netlink_kobject_uevent_socket and read udev runtime…

zpytela:
… files

The permissions are required when rtl-sdr is in place as the entropy source.
The librtlsdr library uses libusb, which uses libudev, which creates and
monitors a NETLINK_KOBJECT_UEVENT socket.

Resolves: rhbz#1818584


wrabcak:
LGTM.

PR #236, opened 2020-04-16
Allow glusterd synchronize between master and slave

zpytela:
Allow communication between glusterd_t and rsync_t using
unix_stream_socket.


zpytela:
Requested a scratchbuild in the meantime to check the resulting permissions set.

wrabcak:
@zpytela  Do you want to wait for results from gluster team or should I merge it?

zpytela:
@wrabcak , I don't think we should wait, I just wanted to doublecheck the results, and I happened to find one oversight. Please merge now.

PR #239, opened 2020-04-27
Update networkmanager_read_pid_files() to allow also list_dir_perms

zpytela:
Update the networkmanager_read_pid_files() interface so that the domain
allowed access can also list the NetworkManager runtime directory.

### DIFF
--- a/networkmanager.if
+++ b/networkmanager.if
@@ -298,6 +298,7 @@ interface(`networkmanager_read_pid_files',`
 	')
 
 	files_search_pids($1)
+	list_dirs_pattern($1, NetworkManager_var_run_t, NetworkManager_var_run_t)
 	read_files_pattern($1, NetworkManager_var_run_t, NetworkManager_var_run_t)
 ')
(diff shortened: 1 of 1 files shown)


zpytela:
Looks like this permission was overlooked when resolving bz#1810404. Needs to be backported to F31 and F30, too.

networkmanager.if
```diff
@@ -298,6 +298,7 @@ interface(`networkmanager_read_pid_files',`
 	')
 
 	files_search_pids($1)
+	allow $1 NetworkManager_var_run_t:dir list_dir_perms;
```
wrabcak: Why not use: list_dirs_pattern?

zpytela:
It was just copied from other interfaces, fixed.

wrabcak (approved):

wrabcak:
No problem, I was just curious.

wrabcak:
Commit backported also to F31 and F30.

PR #240, opened 2020-05-05
Allow nagios_plugin_domain execute programs in bin directories

zpytela:
Allow nagios_plugin_domain execute generic programs in system bin
directories without a domain transition.


wrabcak:
LGTM.

PR #246, opened 2020-05-14
Label /var/run/sympa(/.*)? httpd_var_run_t

ngutlhtej:
As per https://github.com/sympa-community/sympa/issues/799

/var/run/sympa(/.*)? requires httpd_var_run_t SELinux label

This is for the mailing list manager "sympa" currently in Fedora and EPEL.

### DIFF
--- a/apache.fc
+++ b/apache.fc
@@ -180,6 +180,7 @@ ifdef(`distro_debian', `
 /var/run/nginx.*            gen_context(system_u:object_r:httpd_var_run_t,s0)
 /var/opt/rh/rh-nginx18/run/nginx(/.*)?            gen_context(system_u:object_r:httpd_var_run_t,s0)
 /var/run/php-fpm(/.*)?      gen_context(system_u:object_r:httpd_var_run_t,s0)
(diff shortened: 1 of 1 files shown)


zpytela:
@BenjaminLefoul, this commit as such looks good, even a file transition is not needed if the runtime files are only managed by httpd. However, it looks sympa has a few services; without a particular context for the executables, all of the services run in the `unconfined_service_t` domain. Did you think about confining the services to improve security of the package?

In particular, the following permission is unlikely to be allowed in selinux-policy:
```
allow httpd_t unconfined_service_t:unix_stream_socket connectto;
```

ngutlhtej:
Thanks for that quick reply @zpytela 

Yes we need to set the contexts on the executable sympa files.
Note however that that permission is not needed on RHEL8 (which I admit is my main concern right now). I assume this is because RHEL8 has the following permission and Fedora doesn't:

`allow domain unconfined_service_t:unix_stream_socket connectto;`

...but I am not sure, there may be another reason?

PR #249, opened 2020-05-20
Allow rhsmd read process state of all domains and kernel threads

zpytela:
Allow rhsmcertd_t domain_read_all_domains_state() and kernel_read_state().
These permissions are required in newer versions of subscription-manager
that stopped using the psutil python library and implemented its own method
to read /proc.


wrabcak:
LGTM.

PR #251, opened 2020-05-26
Introduce logrotate_use_cifs boolean

zpytela:
The logrotate_use_cifs boolean now can be used to allow logrotate
operate on CIFS volumes.


wrabcak:
LGTM.

PR #252, opened 2020-05-26
httpd: Allow NoNewPriv transition from systemd

travier:
This enables running Apache/nginx directly under an unprivileged user
using systemd with the User/Group directives and optionnaly the
CAP_NET_BIND_SERVICE ambiant capability to bind <1024 ports.

See https://medium.com/@nickodell/sandboxing-nginx-with-systemd-80441923c555 for a recent example.

Tested as a module on Fedora 32.

### DIFF
--- a/apache.te
+++ b/apache.te
@@ -329,6 +329,7 @@ ifdef(`distro_redhat',`
 	typealias httpd_exec_t alias phpfpm_exec_t;
 ')
 init_daemon_domain(httpd_t, httpd_exec_t)
+init_nnp_daemon_domain(httpd_t)
 role system_r types httpd_t;
 
 # httpd_cache_t is the type given to the /var/cache/httpd


apache.te
```diff
@@ -331,6 +331,8 @@ ifdef(`distro_redhat',`
 init_daemon_domain(httpd_t, httpd_exec_t)
 role system_r types httpd_t;
 
+allow init_t httpd_t:process2 nnp_transition;
```
wrabcak: We can use following interface: init_nnp_daemon_domain() for this purpose.

wrabcak (changes requested):

travier:
Updated. Thanks

wrabcak (approved):

wrabcak:
I made some cosmetic details. LGTM. 

Thanks!

travier:
Thanks! Do you think this could be considered for the Fedora 32 branch?

wrabcak:
Hi @travier 
it's already there: https://github.com/fedora-selinux/selinux-policy-contrib/commit/527137880eba5347f1aa1067ec5a60588b86a1ea

travier:
Awesome, thanks!

PR #255, opened 2020-06-01
Allow qemu-kvm read and write /dev/mapper/control

zpytela:
Allow svirt_tcg_t read and write to lvm_control_t char device.


wrabcak:
LGTM

PR #267, opened 2020-06-15
Allow tlp_t execute_no_trans tlp_exec_t

zpytela:
Resolves: rhbz#1844755

### DIFF
--- a/tlp.te
+++ b/tlp.te
@@ -47,6 +47,8 @@ kernel_dontaudit_request_load_module(tlp_t)
 
 auth_read_passwd(tlp_t)
 
+can_exec(tlp_t, tlp_exec_t)
+
 corecmd_exec_bin(tlp_t)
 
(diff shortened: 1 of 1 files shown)


tlp.te
```diff
@@ -27,6 +27,7 @@ allow tlp_t self:unix_stream_socket create_stream_socket_perms;
 allow tlp_t self:udp_socket create_socket_perms;
 allow tlp_t self:unix_dgram_socket create_socket_perms;
 allow tlp_t self:netlink_generic_socket create_socket_perms;
+allow tlp_t tlp_exec_t:file execute_no_trans;
```
wrabcak: Why not to user can_exec() ?

zpytela:
In other policy modules, both approaches are used. In this case, execute_no_trans was the only permission missing. Anyway, I've changed it.

wrabcak:
LGTM

PR #269, opened 2020-06-16
Allow qemu read and write /dev/mapper/control

zpytela:
Allow svirt_t read and write to the lvm_control_t char device.

Resolves: rhbz#1824019


wrabcak:
LGTM

PR #271, opened 2020-06-17
Allow radiusd connect to gssproxy over unix domain stream socket

zpytela:


wrabcak:
Can you please share more info about the PR? Is it expected that there are 2 commits in the PR? Are the commits related?

zpytela:
There are two independent commits, the second one is an outcome of solving the first one and going through the module files.

wrabcak:
OK

PR #278, opened 2020-06-22
Make keepalived work with network namespaces

zpytela:
Assign keepalived_var_run_t type also to directories in /run.
Allow keepalived_t mounton keepalived_var_run_t.


wrabcak:
LGTM.

PR #280, opened 2020-06-24
Allow irqbalance nnp_transition

zpytela:
Allow irqbalance domain transition with NoNewPrivileges

Resolves: rhbz#1848658


wrabcak:
LGTM

PR #286, opened 2020-07-01
Allow irqbalance file transition for pid sock_files and directories

zpytela:
Allow irqbalance file transition to its private type in
runtime directory also for sock_file and directory class.

Resolves: rhbz#1852486

### DIFF
--- a/irqbalance.te
+++ b/irqbalance.te
@@ -26,8 +26,10 @@ dontaudit irqbalance_t self:capability sys_tty_config;
 allow irqbalance_t self:process { getcap getsched setcap signal_perms };
 allow irqbalance_t self:udp_socket create_socket_perms;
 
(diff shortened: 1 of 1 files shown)


wrabcak:
You're allowing to make transition for dirs and sock_files but process irqbalance_t cannot manage them. Is it expected?

zpytela:
It is expected irqbalance be able to manage them, fixed.

wrabcak:
What about dirs?

zpytela:
`manage_files_pattern` already contains `rw_dir_perms` for the directory class, but I see now it may not be enough.

wrabcak:
AFAIK, `rw_dir_perms` is not enough to create new dir. 

LGTM

wrabcak:
btw. make sure that also irqbalance.fc is updated

PR #287, opened 2020-07-01
Update irqbalance runtime directory file context

zpytela:
Update file context specification for irqbalance in the runtime directory
so that it matches /var/run/irqbalance.* for all file types.

Resolves: rhbz#1852486


wrabcak:
LGTM.

PR #292, opened 2020-07-02
[WIP] Add oddjob_noatsecure() interface

zpytela:


zpytela:
An interface for reverse way is required.

PR #294, opened 2020-07-02
[WIP] Allow cupsd access local configuration files

zpytela:
Label ~/.cups with cups_home_t.
Make cups_home_t usable in a user home directory.
Allow cupsd_t read files with the cups_home_t type.

Resolves: rhbz#1838495


zpytela:
Updated. See also https://github.com/fedora-selinux/selinux-policy/pull/390
Need to test it together first, then request review again.

PR #297, opened 2020-07-03
Allow certmonger manage dirsrv services

zpytela:


zpytela:
Another mysterious travis message:
```
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```
and no button to restart CI.

Applying in a scratch build together with `0001-Run-ipa_helper_noatsecure-oddjob_t-only-if-the-inter.patch` it works.
https://koji.fedoraproject.org/koji/taskinfo?taskID=46526888

WOnder93:
> and no button to restart CI.

I couldn't see it either and Travis was telling me I'm not a member of any organization... Somehow when I clicked my avatar in the upper right corner and then "Add an account" -> "Sign in with GitHub", then it got fixed. Maybe just signing out & in would also do the trick.

wrabcak:
LGTM.

PR #303, opened 2020-07-08
Remove systemd_dbus_chat_resolved(pcp_pmie_t)

zpytela:
The permissions are now allowed for the domain attribute.


zpytela:
Requires https://github.com/fedora-selinux/selinux-policy/pull/394 to be merged.

PR #308, opened 2020-07-23
Allow munin domain transition with NoNewPrivileges

zpytela:
Resolves: rhbz#1857381


wrabcak:
LGTM. Backporting also to F32 and F31

PR #312, opened 2020-07-27
Label /tmp/krb5_0.rcache2 with krb5_host_rcache_t

zpytela:
Label /tmp/krb5_0.rcache2 with krb5_host_rcache_t,
define named file transition for kadmind on /tmp/krb5_0.rcache2.


zpytela:
I've just realized smbd uses the krb library instead of kadmind, so need to have the transition, too.

PR #314, opened 2020-07-29
Allow virtlockd manage VMs posix file locks

zpytela:
Label /var/lib/libvirt/lockd with the new virt_var_lockd_t type.
Allow virtlockd manage virt_var_lockd_t files and directories.
Create the virt_lock_blk_devs boolean to allow virtlockd read/lock
block device nodes.


wrabcak:
LGTM.

PR #315, opened 2020-07-30
squid: Allow net_raw capability when squid_use_tproxy is enabled

romanofski:
When the SELinux boolean `squid_use_tproxy` is enabled, this module
allows Squid net_admin capabilities. However `net_raw` will be denied.
The capability however is needed when squid acts as a transparent proxy
in circumstances also outlined in the capabilites(7) man page:

    CAP_NET_RAW

       * Use RAW and PACKET sockets;
       * bind to any address for transparent proxying.

This patch adds `net_raw` to the capabilities which will be allowed if
`squid_use_tproxy` is enabled.

### DIFF
--- a/squid.te
+++ b/squid.te
@@ -200,7 +200,7 @@ tunable_policy(`squid_connect_any',`
 ')
 
 tunable_policy(`squid_use_tproxy',`
-	allow squid_t self:capability net_admin;
+	allow squid_t self:capability { net_admin net_raw };
 	corenet_sendrecv_netport_server_packets(squid_t)
 	corenet_tcp_bind_netport_port(squid_t)
 	corenet_tcp_sendrecv_netport_port(squid_t)


romanofski:
PS: In case this patch is acceptable, don't merge right away. I'd like to compile and test first.

zpytela:
@romanofski, could you share the current status of this PR? The content looks reasonable to me, but you mentioned some testing.

romanofski:
Dear @zpytela sorry for the late reply and the missing follow up comment. When I implemented I checked that I didn't run into any regressions with the system under test. I don't have access to the original system any more, but would think the patch's impact to be small.

zpytela:
@romanofski to say it clear: this PR is ready to merge, right?

romanofski:
@zpytela yes. Merge ho!

zpytela:
@romanofski thank you, merging.

PR #317, opened 2020-07-30
Allow qemu-ga read all non security file types conditionally

zpytela:
Create the virt_qemu_ga_read_nonsecurity_files boolean to conditionally
allow qemu guest agent read all directories, files, and symlinks of the
types which are a part of the non_security_file_type attribute.
This is required to do trim and freeze at any mountpoint.


wrabcak:
LGTM

elmarco:
thanks

PR #321, opened 2020-08-07
Label /usr/libexec/qemu-pr-helper with virtd_exec_t

zpytela:
Label /usr/libexec/qemu-pr-helper with the virtd_exec_t type.
The file path has moved in qemu-kvm-5.0.0 so that now it matches
the /usr/libexec/qemu.* regexp and gets an incorrect type.


wrabcak:
LGTM.

berrange:
At some point we should figure out a custom poliucy for qemu-pr-helper, as there's no way it needs the full range of permissions allowed by virtd_t. Using virtd_t makes it little better than being unconfined.  qemu-pr-helper has such a small range of functionality we can do much better at lockdown.

PR #322, opened 2020-08-12
Allow check_multipath work in nrpe

zpytela:


zpytela:
The previous commit allowed the permissions to `nrpe_t` which is not necessary when the plugin has appropriate type.

Dbus chat is required when sudo is in place. Using sudo is for sure required for `nagios_system_plugin_t`, `nagios_mail_plugin_exec_t`, and `nagios_unconfined_plugin_exec_t`, so allowing to all plugins, conditionally.

wrabcak:
LGTM.

PR #327, opened 2020-08-31
Allow plymouth sys_chroot capability

zpytela:
This capability is required as chroot() is called in the plymouth source code.


wrabcak:
LGTM

PR #330, opened 2020-09-02
Allow journalctl to read and write to inherited user domain tty

Koncpa:
Add macro userdom_use_inherited_user_tty() to journalctl policy, which
allow to read and write to inherited user domain tty.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1767721


wrabcak:
LGTM.

PR #337, opened 2020-10-05
Allow cups_pdf_t domain to communicate with unix_dgram_socket

Richard-Filo:
Allow cups_pdf_t domain to communicate with processes labeled cups_pdf_t on same machine using unix_dgram_socket.
The cups-pdf service had allowed communication with unix_stream_socket but now the service starts to use another type of socket.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1832521

### DIFF
--- a/cups.te
+++ b/cups.te
@@ -614,6 +614,7 @@ optional_policy(`
 #
 
 allow cups_pdf_t self:capability { chown fowner fsetid setuid setgid dac_read_search dac_override };
+allow cups_pdf_t self:unix_dgram_socket create_stream_socket_perms;
 allow cups_pdf_t self:unix_stream_socket create_stream_socket_perms;
 allow cups_pdf_t cupsd_rw_etc_t:dir search;
 


Richard-Filo:
```
# sesearch -A -s cups_pdf_t -c unix_dgram_socket

```
after installation of scratch build:
```
# sesearch -A -s cups_pdf_t -c unix_dgram_socket
allow cups_pdf_t cups_pdf_t:unix_dgram_socket { accept append bind connect create getattr getopt ioctl listen lock read setattr setopt shutdown write };
```

zpytela:
Although the rule for keeping alphabetial order is not always followed, I'd rather see lines 617 and 618 swapped. Also please modify the first line of the commit message to be more concise - yes, another rule, although we don't insist on 50 characters limit, just to have it readable: see `git log` and `git log --oneline` to see the message in context.

Richard-Filo:
@zpytela I hope it is better now.

zpytela:
Looks better. Surely you could have kept the long version in the committ message body.

zpytela:
@Richard-Filo I don't want to be nitpicking only, but can also the commit body have wrapped lines? Again, 72 characters is not a hard limit, rather recommendation worth following. Refer to `git log` for the readability context.

Richard-Filo:
@zpytela is it ok now ?

zpytela:
LGTM, thank you.

PR #339, opened 2020-10-08
rpc.fc: Include /etc/exports.d dir & files

tasleson:
The directory /etc/exports.d and files which reside in it should
have the exports_t context.
    
see: man 5 exports
    
Signed-off-by: Tony Asleson <tasleson@redhat.com>

### DIFF
--- a/rpc.fc
+++ b/rpc.fc
@@ -4,6 +4,7 @@
 # /etc
 #
 /etc/exports		--	gen_context(system_u:object_r:exports_t,s0)
+/etc/exports\.d(/.*)?		gen_context(system_u:object_r:exports_t,s0)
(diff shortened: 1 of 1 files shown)


rpc.fc
```diff
@@ -4,6 +4,8 @@
 # /etc
 #
 /etc/exports		--	gen_context(system_u:object_r:exports_t,s0)
+/etc/exports\.d		--	gen_context(system_u:object_r:exports_t,s0)
```
zpytela: The `--` file specifier means a plain file, so this pattern would never apply.

zpytela (commented):
You are right, I think we can assign the same type to any file though.

zpytela:
LGTM, thanks.

PR #344, opened 2020-10-14
Create chronyd_pid_filetrans() interface

zpytela:


zpytela:
Merging and backporting to F33.

PR #347, opened 2020-10-20
Allow openvswitch fowner capability and create netlink sockets

zpytela:
Resolves: rhbz#1844198


zpytela:
CI fails as working with the proper branch has not been fixed for contrib.

PR #348, opened 2020-10-23
Allow slapd add new entries in certificates directories

zpytela:


zpytela:
This commit effectively replaces
```
define(`list_dir_perms',`{ getattr search open read lock ioctl }')
```
with
```
define(`add_entry_dirs_pattern',`
        allow $1 $2:dir search_dir_perms;
        allow $1 $3:dir add_entry_dir_perms;
')
define(`search_dir_perms',`{ getattr search open }')
define(`add_entry_dir_perms',`{ getattr search open lock ioctl write add_name }')
```

zpytela:
Requires a new interface from https://github.com/fedora-selinux/selinux-policy/pull/465

WOnder93 (approved):

zpytela:
The build fails as it had started before the interface PR was merged.

certmonger.te
```diff
@@ -101,6 +101,7 @@ libs_exec_ldconfig(certmonger_t)
 
 logging_send_syslog_msg(certmonger_t)
 
+miscfiles_add_entry_generic_cert_dirs(certmonger_t)
 miscfiles_manage_all_certs(certmonger_t)
```
WOnder93, line 105: Now that I looked at this again, isn't the newly added interface call a subset of this one (`miscfiles_manage_all_certs(certmonger_t)`)? (But judging from #467 you might have already noticed, too :)

zpytela: It is a subset; however, I did not want to allow it as manage allows the domain also remove-related permissions. Thinking about it again, at least rename will be needed as the next step.

WOnder93: Well, since the created directory from the BZ is named `dbTemp.Vmrrvi` and this seems to be some sort of cache, I'd expect that the domain might want to also remove the dir at some point so even adding `remove` will probably make sense.

One way or another, you should at least remove `miscfiles_add_entry_generic_cert_dirs()` (both the call and definition), which is just redundant.

zpytela: I haven't realized remove_name is a part of the process, now it makes sense.

Not important any longer, but IMHO there are no redundancies.

WOnder93: `miscfiles_add_entry_generic_cert_dirs(certmonger_t)` is a subset of `miscfiles_manage_all_certs(certmonger_t)` and this is the only caller. So unless you plan to remove `miscfiles_manage_all_certs(certmonger_t)`, it *is* redundant.
```
$ macro-expander 'miscfiles_manage_all_certs(certmonger_t)'
allow certmonger_t cert_type:dir { getattr search open read lock ioctl };
allow certmonger_t cert_type:dir { open read getattr lock search ioctl add_name remove_name write };
allow certmonger_t cert_type:file { create open getattr setattr read write append rename link unlink ioctl lock };
allow certmonger_t cert_type:dir { open read getattr lock search ioctl add_name remove_name write };
allow certmonger_t cert_type:lnk_file { create getattr setattr read write append rename link unlink ioctl lock };
$ macro-expander 'add_entry_dirs_pattern(certmonger_t, cert_t, cert_t)'
allow certmonger_t cert_t:dir { getattr search open };
allow certmonger_t cert_t:dir { getattr search open lock ioctl write add_name };
$ seinfo -xt cert_t

Types: 1
   type cert_t, file_type, non_auth_file_type, non_security_file_type, cert_type;
```

PR #349, opened 2020-10-23
Add default file context for /usr/libexec/pcp/lib/*

zpytela:
As a result of changes in pcp systemd units, some executables were moved
to /usr/libexec/pcp/lib. The default file context specification in the
selinux policy needs to be adjusted accordingly.


natoscott:
LGTM.

PR #350, opened 2020-10-26
Remove retired PCP pmwebd and pmmgr daemons (since 5.0)

natoscott:


zpytela:
@natoscott thank you for the PR. It looks v5.0 has been in Fedora for a while (1 year), so if the two services do not exist any longer, we can remove unneeded permissions. I will also backport it to F33.

natoscott:
@zpytela yep, its safe to drop these now - these daemons are long gone from Fedora.

zpytela:
F33 backport: https://github.com/fedora-selinux/selinux-policy-contrib/pull/353

PR #351, opened 2020-10-27
Let keepalived open a raw socket

mscherer:
See https://bugzilla.redhat.com/show_bug.cgi?id=1891895

### DIFF
--- a/keepalived.te
+++ b/keepalived.te
@@ -62,6 +62,7 @@ auth_use_nsswitch(keepalived_t)
 corecmd_exec_bin(keepalived_t)
 corecmd_exec_shell(keepalived_t)
 
+corenet_raw_bind_generic_node(keepalived_t)
 corenet_tcp_connect_connlcli_port(keepalived_t)
 corenet_tcp_connect_http_port(keepalived_t)
 corenet_tcp_connect_mysqld_port(keepalived_t)


keepalived.te
```diff
@@ -43,6 +43,7 @@ allow keepalived_t self:netlink_connector_socket create_socket_perms;
 allow keepalived_t self:netlink_route_socket nlmsg_write;
 allow keepalived_t self:packet_socket create_socket_perms;
 allow keepalived_t self:rawip_socket create_socket_perms;
+corenet_raw_bind_generic_node(keepalived_t)
```
zpytela: @mscherer could you move the line to the other corenet_ calls?

zpytela (changes requested):
I think it should rather be binding than opening.

keepalived.te
mscherer: yup

zpytela:
I tried to look up rules for ordering within corenet interfaces block usage, without success though. It seems to me the order is not alphabetical as it is in most of the cases, but rather all-generic-particular. Having said that, I'd rather move it at the block beginning.

zpytela (approved):

zpytela:
@mscherer, thanks for the excellent cooperation.

zpytela:
Cherry-picked for rawhide, too: https://github.com/fedora-selinux/selinux-policy-contrib/pull/352

PR #358, opened 2020-11-03
Introduce rpmdb_t type

zpytela:
When a rpm database is rebuilt using the rpmdb command, a temporary
directory in /var/lib is created to be renamed to rpm later.
This requires rpmdb run in a private domain so that an unnamed
transition can be defined.

This commit adds rpmdb_t and rpmdb_exec_t types, rpm_domtrans_rpmdb()
and rpm_run_rpmdb() interfaces.

Resolves: rhbz#1461313

### DIFF
--- a/rpm.fc
+++ b/rpm.fc
@@ -8,6 +8,7 @@
 /usr/bin/dnf-automatic  --  gen_context(system_u:object_r:rpm_exec_t,s0)
 /usr/bin/dnf-[0-9]+		--	gen_context(system_u:object_r:rpm_exec_t,s0)
 /usr/bin/rpm 			--	gen_context(system_u:object_r:rpm_exec_t,s0)
+/usr/bin/rpmdb 			--	gen_context(system_u:object_r:rpmdb_exec_t,s0)
 /usr/bin/smart 			--	gen_context(system_u:object_r:rpm_exec_t,s0)
 
 /bin/yum-builddep		--	gen_context(system_u:object_r:rpm_exec_t,s0)
--- a/rpm.if
+++ b/rpm.if
@@ -115,6 +115,50 @@ interface(`rpm_exec',`
 
 ########################################
 ## <summary>
+##	Execute rpmdb in the rpmdb domain.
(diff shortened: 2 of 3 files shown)


zpytela:
The database rebuild verified working for unconfined root and sysadm_u.

WOnder93:
I'm not sure if confining `rpmbdb` for unconfined_t is a good idea... `rpmdb` can be used for more than just DB rebuild. For example, these options could be a concern and \<speculation\>might not work after this change\</speculation\>:
```
Usage: rpmdb [OPTION...]
[...]
      --exportdb                 export database to stdout header list
      --importdb                 import database from stdin header list
[...]
      --load=<FILE>              load a single macro file
[...]
      --rcfile=<FILE:...>        read <FILE:...> instead of default file(s)
  -r, --root=ROOT                use ROOT as top level directory (default: "/")
      --dbpath=DIRECTORY         use database in DIRECTORY
[...]
```
Depending on where the user points the filenames, it may happen that `rpmdb_t` won't have the permissions to access those locations (while `unconfined_t` would).

I suppose it could be fixed by having a different `rpmdb_unconfined_t` type, to which only `unconfined_t ` would transition, but that could be quite messy... Yet, I don't have any better solution to offer that could be implemented quickly (other than adding SELinux-specific code to `rpmdb`) :(  I really should finally get around to adding regex/glob support to filename transitions...

zpytela:
I see confining of rpmdb as the best approach given the current options we have.

If accessing other locations is the only problem, we can allow accessing non_security_file_type, analogously to chronyc_t which is also confined cli command.

WOnder93:
Ok, it's your call.

zpytela:
@WOnder93 thank you for your reviews which make the results better, I am grateful for any comment.

Just verified read/write is enough to write to /root and read from almost anywhere.
```
(allow rpmdb_t non_security_file_type (file ( read write )))
```

zpytela:
Merging per positive results of internal testing.

PR #361, opened 2020-11-09
Introduce rpmdb_t type

zpytela:
When a rpm database is rebuilt using the rpmdb command, a temporary
directory in /var/lib is created to be renamed to rpm later.
This requires rpmdb run in a private domain so that an unnamed
transition can be defined.

This commit adds rpmdb_t and rpmdb_exec_t types,
rpm_domtrans_rpmdb() and rpm_run_rpmdb() interfaces.
Additionally, read/write access to non-security files is allowed to
provide a way for rpmdb to use files when stdin or stdout is expected.

Resolves: rhbz#1461313


zpytela:
The build fails since chronyd_pid_filetrans has not been backported to f32 and contrib from f33 branch is used.

PR #364, opened 2020-11-13
Change fetchmail temporary files path to /var/spool/mail

zpytela:
Until now, the .fetchmail-UIDL-cache and .fetchmail.pid files had
default file context defined for the /var/mail path, in Fedora though
/var/mail is a symlink to spool/mail, so the path actually needs to be
/var/spool/mail.

Resolves: rhbz#1840880


zpytela:
Tested.

PR #365, opened 2020-11-14
Add knot policy

mscherer:


mscherer:
Ideally, once that's merged, I would love to have it on F33 (as that's what I used to test, and I plan to deploy knot in prod once there is some selinux policy)

zpytela:
@mscherer, looks good at a first glance, need to check it deeper though.

How did you test the policy? There seems to be a test suite as a part of the package.

mscherer:
I just installed the server and operated it, and remove AVC as they came. I will take a look at the test suite, and I also need to exercise the plugin and others things (remote control execution, etc). Maybe I should tag it as WIP in fact.

mscherer:
Looking at the tests, it seems the one in Fedora are just "add a default zone, and test if it answer". That's a bit basic for our purpose, and I think a deployment in production would cover that part. The tests in the knot tarball are written in C and from  a quick look, it seems to be mostly unit tests for the code base, so I doubt that can be used for testing the policy. There isn't a clear separation between the test code and knot code since all would be done in the same process and it doesn't seems most of the test touch to real life locations/ports/etc.

So I am just gonna compile my policy, run it for 1 or 2 weeks, fix, repeat, etc and come back unless I forget about it.

mscherer:
I checked the whole documentation, the only feature that could requires some change is support for certificates on PKCS11 devices. And I do not have the required hardware to test, even if it can also be done with softhsm (where I do not have the experience, nor know how it work.

It seems bind also already deal with that:
```
/var/lib/softhsm(/.*)?      gen_context(system_u:object_r:named_cache_t,s0)
```

But bind.fc seems also cover unbound, which is surprising since nsd is in a separate policy.

mscherer:
Closed in favor of https://github.com/fedora-selinux/selinux-policy/pull/539

PR #366, opened 2020-11-16
Update apache file context for /etc/httpd symlinks to get etc_t type

zpytela:
For compatibility across different systems, apache httpd has the default
file context specification for symlinks in /etc/httpd to different
filesystem directories (state directory, runtime files, library modules,
logs, etc.) The symlinks in /etc/httpd, like other files without a private
type, have the default httpd_config_t. This setting can prevent domains
without the permission to read httpd_config_t from working, while all
domains have access to symlinks with the etc_t type.

Resolves: rhbz#1890024


zpytela:
Tested.

PR #367, opened 2020-11-16
Allow dnsmasq read public files

zpytela:
As a part of dnsmasq code, there is also an implementation of a tftp
server. Read access to public content files used for file transfer
services is required for dnsmasq to be able to provide the tftp service.

Resolves: rhbz#1782539


zpytela:
Tested.

PR #368, opened 2020-11-19
Allow rpmdb rw access to inherited console, ttys, and ptys

zpytela:
With introducing the rpmdb_t private type, rpmdb also needs read and write
inherited console, all tty, and all pty terminals to be able to display
messages, e. g. when --help switch is used.

Resolves: rhbz#1899548


zpytela:
Tested.

PR #371, opened 2020-11-23
Allow cups_pdf_t domain to communicate with unix_dgram_socket

zpytela:
Allow cups_pdf_t domain to communicate with processes labeled
cups_pdf_t on same machine using unix_dgram_socket.The cups-pdf
service had allowed communication with unix_stream_socket but now
the service starts to use another type of socket.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1832521


zpytela:
Fails because of branches mismatch, otherwise looks good and the build builds.

PR #379, opened 2020-12-01
F32 targetd lvm

zpytela:


zpytela:
Build fails because of repos branches mismatch.

PR #383, opened 2020-12-07
Allow dovecot bind to smtp ports

zpytela:
When dovecot is configured to listen on submission ports
(tcp 465 or 587), it requires the name_bind permission to ports
labeled smtp_port_t.

Resolves: rhbz#1889142


zpytela:
Fails because of branches mismatch, the PR content is the same as in https://github.com/fedora-selinux/selinux-policy-contrib/pull/382 which succeeded.

PR #386, opened 2020-12-15
Allow dovecot_auth_t stat /proc filesystem

zpytela:


WOnder93:
Should https://github.com/fedora-selinux/selinux-policy/commit/785340ee2d313990024653ce2dd028d34bdeecd9 also be backported? IIUC the root cause is the same so I'd expect the same problems to occur on F33. Or am I missing something?

zpytela:
@WOnder93 in F33 there is libcap-ng-0.8

WOnder93:
But didn't your analysis show that libpam also does the statfs directly? (Anyway, the patch itself is OK, so I'll give it the green tick. The rest can be added later if really needed.)

WOnder93 (approved):

zpytela:
I'd like to leave it as is, but surely will be aware of changes possibly required in the future.

PR #387, opened 2021-01-11
Allow sssd read /run/systemd directory

zpytela:
The nsswitch_domain is already allowed search /run/systemd, sssd however
requires the read permission, granted by the list_dir_perms pattern.

The reason is that sssd is using an asynchronous resolver library
(c-ares) and monitors /etc/resolv.conf for changes. If /etc/resolv.conf
is replaced with a symlink, SSSD tries to follow it to set an inotify
watch to be aware of the target file changes. The resolv.conf file
changes can be made by a user, NetworkManager, or systemd-resolved.

Resolves: rhbz#1827466

### DIFF
--- a/sssd.te
+++ b/sssd.te
@@ -129,6 +129,8 @@ auth_manage_cache(sssd_t)
 auth_login_manage_key(sssd_t)
 dontaudit sssd_t init_t:dbus send_msg;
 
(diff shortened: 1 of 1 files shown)


zpytela:
CI fails due to a known problem with mismatched repos. Merging per successful F33 CI:
https://github.com/fedora-selinux/selinux-policy-contrib/pull/388

PR #390, opened 2021-03-17
Dontaudit NetworkManager write to initrc_tmp_t pipes

zpytela:
Resolves: rhbz#1892556


zpytela:
F33 backport of https://github.com/fedora-selinux/selinux-policy/pull/604

zpytela:
CI fails because it builds with latest rawhide policy, otherwise it looks good, so merging.

PR #391, opened 2021-03-18
Allow confined users login into graphic session

zpytela:
Allow dbus role as (user_dbusd_t, staff_dbusd_t and etc.) start systemd services,
modify the systemd configuration of any file and reload the services.
Allow the gkeyringd role as (user_gkeyringd_t, staff_gkeyringd_t and etc.)
set the schedule on self process. Also fix wrong tab space in gnome.if.

allow $1_gkeyringd_t self:process setsched;

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1767874
Migrated PR from selinux-policy-contrib.


zpytela:
CI build fails as it uses the latest policy, otherwise the PR looks good.
Backported from https://github.com/fedora-selinux/selinux-policy/pull/544

PR #392, opened 2021-03-25
Allow bluetooth_t dbus chat with fwupd_t

zpytela:
Resolves: rhbz#1939238


zpytela:
Merging as this is only a backport of a verified commit, regardless of the CI build failure because of mismatching repos.

PR #393, opened 2021-03-25
Allow postgrey read network state

zpytela:
Resolves: rhbz#1943136


zpytela:
Merging as this is only a backport of a verified commit, regardless of the CI build failure because of mismatching repos.

PR #394, opened 2021-04-09
Allow brltty create and use bluetooth_socket

zpytela:
Resolves: rhbz#1946020


zpytela:
Build has failed because of branches mismatch. Merging as this is just a clear backport.

PR #395, opened 2021-04-09
Add default file context for "/var/run/chrony-dhcp(/.*)?"

zpytela:
With the 25d2a5c01c commit, a file transition for /var/run/chrony-dhcp
was added, but it was not backed by the default file context
specification.

Resolves: rhbz#1880948


zpytela:
Merging a cherry-picked commit, the CI error is expected.

PR #396, opened 2021-04-26
Allow IPsec and Certmonger to use opencryptoki services

zpytela:
Add to certmonger and ipsec policy interface pkcs_use_opencryptoki(),
which allow use opencryptoki. Opencryptoki implements PKCS#11
standard.

Resolves: rhbz#1952311


zpytela:
Merging cherry-picked commit which has been confirmed working.

PR #398, opened 2021-05-13
Add permissions to delete lnk_files into gnome_delete_home_config()

zpytela:
The gnome_delete_home_config() interface contains delete_files_pattern()
call for config_home_t files only, but symlinks can be there, too.

Resolves: rhbz#1909344


zpytela:
Merging cherry-picked commit.

PR #400, opened 2021-05-17
Label /var/lib/kdump with kdump_var_lib_t

zpytela:
The kexec-tools usually create initramfs-KERNELVERSIONkdump.img in /boot.
On some operating systems, the /boot directory can be read-only,
preventing the image file from being created. In newer kexec-tools
package versions, the image files are created in /var/lib/kdump
in case /boot is read-only.

The kdump_manage_lib_files() interface was created and rpm_script_t
allowed to manage /var/lib/kdump files. This is needed when the
/lib/kernel/install.d/60-kdump.install kernel install hook is triggered
to delete unused images.

Resolves: rhbz#1951323

### DIFF
--- a/kdump.fc
+++ b/kdump.fc
@@ -13,4 +13,6 @@
 
 /var/crash(/.*)?		gen_context(system_u:object_r:kdump_crash_t,s0)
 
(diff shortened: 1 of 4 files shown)


zpytela:
Merging a cherry-picked commit.

